PR #3, opened 2024-02-02
Image-level metadata extraction

empurvis:
Image-level metadata extraction functions and wrapper

### DIFF
--- a/sandbox/extract-imagery-metadata_emp.R
+++ b/sandbox/extract-imagery-metadata_emp.R
@@ -127,10 +127,7 @@ dataset_id
 
 # BEGIN FUNCTION CODE
 
-exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal)
-exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
-exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
-exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
+exif$DateTimeOriginal_nocolon = stringr::str_replace_all(exif$DateTimeOriginal, ":", "")
 
 datatime_local = (exif$DateTimeOriginal_nocolon)
 
@@ -140,10 +137,7 @@ datatime_local = (exif$DateTimeOriginal_nocolon)
 
 extract_datatime_local = function (exif) {
 
-  exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal)
-  exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
-  exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
-  exif$DateTimeOriginal_nocolon <- sub(":", "", exif$DateTimeOriginal_nocolon)
+  exif$DateTimeOriginal_nocolon = stringr::str_replace_all(exif$DateTimeOriginal, ":", "")
 
   datatime_local = (exif$DateTimeOriginal_nocolon)
 
@@ -386,9 +380,11 @@ white_balance
 
 # BEGIN FUNCTION CODE
 
-received_image_path = stringr::str_split_fixed(exif$SourceFile, "/", 5)
+received_image_path = stringr::str_split_fixed(exif$SourceFile, fixed(dataset_id), 2)
 
-received_image_path <- received_image_path[,5]
+received_image_path <- received_image_path[,2]
+
+received_image_path <- with(exif, paste0(dataset_id, received_image_path))
 
 # END FUNCTION CODE
 
@@ -396,9 +392,11 @@ received_image_path <- received_image_path[,5]
 
 extract_received_image_path = function(exif) {
 
-  received_image_path = stringr::str_split_fixed(exif$SourceFile, "/", 5)
+  received_image_path = stringr::str_split_fixed(exif$SourceFile, fixed(dataset_id), 2)
 
-  received_image_path <- received_image_path[,5]
+  received_image_path <- received_image_path[,2]
+
+  received_image_path <- with(exif, paste0(dataset_id, received_image_path))
 
   return(received_image_path)
 }
--- a/sandbox/extract-imagery-metadata_emp.R
+++ b/sandbox/extract-imagery-metadata_emp.R
@@ -1,0 +1,578 @@
+# This script has two purposes: 1. Demonstrate how to run functions to extract metadata from imagery
+# dataset EXIF data, and 2. Demonstrate how to write a new metadata extraction function.
+
+# --- Setup ---
+
+# Load all the functions (and package dependencies) of this R package
+devtools::load_all()
+
+# Define the root of the local data directory
+# datadir = readLines(file.path("sandbox", "data-dirs", "derek-metadata-laptop.txt"))
+
+datadir = readLines(file.path("sandbox", "data-dirs", "emp-metadata-laptop.txt"))
+
+
+# --- 1. Workflow for running metadata extraction ---
+
+# Get a list of the files containing the test EXIF data (one file per image dataset). These files
+# have already been created and saved into the project data folder.
+exif_files = list.files(file.path(datadir, "exif-examples"), pattern = "^exif.+\\.csv$", full.names = TRUE)
+
+# Define which test EXIF file to run the functions on
+exif_file = exif_files[1]
+
+# Run for that one EXIF file.
+extract_metadata_dy(exif_file, plot_flightpath = TRUE)
+# ^ If you want to inspect the definition of this function, it is at the bottom of
+# 'R/imagery-metadata-extraction_dy.R'.
+
+# Run extraction on all EXIF files
+metadata = purrr::map_dfr(exif_files, extract_metadata_dy, plot_flightpath = TRUE)
+metadata
+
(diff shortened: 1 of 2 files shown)


sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# BEGIN FUNCTION CODE
+
+datatime_local = (exif$DateTimeOriginal)
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_datatime_local = function (exif) {
```
youngdjn, line 138: This returns it in this format: `2022:06:30 14:32:06` but I would prefer it without the colons. Sorry for the annoyance!

sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_rtk_fix = function(exif) {
+
+  exif["RtkFlag"[!("RtkFlag" %in% colnames(exif))]] = NA
+
+  rtk_fix = ifelse (exif$RtkFlag == 50, "TRUE", ifelse (exif$RtkFlag == 0, "FALSE", "Other"))
```
youngdjn: We don't need to return `"Other"`. I think it's OK to assume that if it's not `50` or the column is missing, it's `FALSE`.
Also, if `RtkFlag` is missing, let's return `FALSE`.
I think I'd also prefer to return a true boolean object, i.e. `FALSE` instead of `"FALSE"`. 

Here's a new way I'd write the body of this function to align with the strategy of dealing with brands (or models) that don't have the right column. (Please check that it runs right and tweak if necessary!)

```R
if ("RtkFlag %in% names(exif)) {
  rtk_fix = exif$RtkFlag == 50
  return(rtk_fix)
} else {
  rtk_fix = rep(FALSE, nrow(exif))
  return(rtk_fix)
}
```

This allows us to add another `else if` block if/when we become aware of a drone model that has a different name/values for this column.

sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# BEGIN FUNCTION CODE
+
+accuracy_x = sqrt(as.numeric(exif$RtkStdLon))
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_accuracy_x = function (exif) {
```
youngdjn: I know what you did is what I requested, but on seeing this I think we should merge the x, y, z accuracy extraction into a single function since the approach should be identical for the three attributes (within a given drone/camera make). So would you mind changing this into a single function that returns a 3-column matrix with column names `accuracy_x`, etc...? Same actually goes for lat/lon and roll/pitch/yaw values (combine lat & lon into a joint function etc). This will also require an update to the wrapper function to accept this new multi-column returned value. If the extraction function returns a matrix (or even data frame) with column names, I think it should work for the wrapper function to build the data frame like your're doing but with:

```
 metadata = data.frame(dataset_id = dataset_id,
                        datatime_local = datatime_local,
                        lat_lon,
                        rtk_fix = rtk_fix,
                        accuracy,
                        pitch_roll_yaw,
                        exposure = exposure,
                        aperture = aperture,
                       ...                
```
(the key above being that the multi-column values are passed to the data.frame function without names, so that they retain the names they already have)

sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# BEGIN FUNCTION CODE
+
+camera_pitch = exif$GimbalPitchDegree
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_camera_pitch = function(exif) {
```
youngdjn: Can you please change these to report pitch, roll, yaw relative to the reference I commented in the protocol doc? E.g., pitch should be degrees up from nadir. (Currently a nadir pitch is reported as `-90`).

Also (sorry), I think the pitch, roll, yaw functions should be merged into a single function for the same reasoning as accuracy.

sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# BEGIN FUNCTION CODE
+
+white_balance = ifelse (exif$WhiteBalance == 0, "Automatic", ifelse (exif$WhiteBalance == 1, "Manual", "Unknown"))
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_white_balance = function(exif) {
```
youngdjn, line 364: What do you think about writing this using `dplyr::case_when` for better readability and easier future extension?
Let's also have the returned values exactly match what's specified in the protocol doc, so `auto` and `manual` instead of `Automatic` etc.

sandbox/extract-imagery-metadata_emp.R
```diff
@@ -0,0 +1,591 @@
...
+
+# BEGIN FUNCTION CODE
+
+received_image_path = exif$SourceFile
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_received_image_path = function(exif) {
```
youngdjn, line 393: This needs a tweak so that the top folder of the returned path is the folder just beneath the dataset folder (which is named like `20220630-0041`). We don't want to see the folders above that.

The best way to achieve this is unclear. One idea: use `stringr::str_split` using the split string `fixed(dataset_id)`, and take the second element (the part that comes after the dataset_id).

youngdjn (changes requested):
Woohoo, it's great to see a working prototype! I requested a few tweaks in specific sections before we merge this. I'd appreciate if you make these changes in this same branch, as that will automatically update this pull request so that we can merge it when the code is ready.

In reviewing this I realized I'd like to add two more columns to the extraction: altitude_atl and altitude_asl (above takeoff level and above sea level, respectively). They correspond to EXIF attributes AbsoluteAltitude and RelativeAltitude. Could you please add these to the extraction protocol doc and the extraction code? Thank you!

You can also delete the parts of my example code (most/all of it?) that is not needed for your code to run.

sandbox/extract-imagery-metadata_emp.R
empurvis: great idea! done

sandbox/extract-imagery-metadata_emp.R
empurvis: done! pitch was the only one I needed to change

sandbox/extract-imagery-metadata_emp.R
empurvis: All good points. I originally put in "other" because there are a bunch DJI values that can be returned in this field. Here are some of those: 
0 - Failed to position 
16 - Single point positioning (meter-level accuracy) 
32~49 - Floating point solution positioning (decimeter-level to meter-level accuracy) 
50 - Fixed solution positioning (centimeter-level accuracy)

So of course 50 is true and 0 (or a missing column) is false, but what about the meter to decimeter level accuracy? Happy to add in other values from this list that return true or false, let me know!

sandbox/extract-imagery-metadata_emp.R
empurvis: done

sandbox/extract-imagery-metadata_emp.R
empurvis: done

sandbox/extract-imagery-metadata_emp.R
youngdjn: 50 (fixed) is the only one that is true RTK-level precision; I think we are best off assuming all the rest are of equivalent "non-RTK presicion". So let's go wtih 50 being TRUE and anything else (including missing) being FALSE -- until we learn of a drone that uses a different code for "fixed".

sandbox/extract-imagery-metadata_emp.R
youngdjn: Could you please use `stringr::str_replace_all` for this, so that it just requires one line?

sandbox/extract-imagery-metadata_emp.R
youngdjn: This will work for this case (imagery folder is 5 levels below top level), but not for a case where the imagery dataset is embedded deeper in more nested folders (or shallower). I'd suggest replacing `"/"` with `fixed(dataset_id)` and taking element 2.

youngdjn (changes requested):
Awesome, this is great! I really like the multi-value extraction functions (lat_lon etc) -- really helps to streamline it. I just have two annoying followup comments, and after that it should be good to merge!

sandbox/extract-imagery-metadata_emp.R
youngdjn: Looks great

sandbox/extract-imagery-metadata_emp.R
youngdjn: Which I see is what you did --thanks!

sandbox/extract-imagery-metadata_emp.R
empurvis: This removes the dataset_id level too, which I don't think you want based on the description in the google doc. I can change it to the level above that (the folder called "1_manually-cleaned), but I think you'll still run into the same problem if and when you have data in a different folder structure. Do you prefer to 1. drop the dataset id folder or 2. keep the dataset id folder but possibly run into problems down the line if some datasets don't have a "1_manually-cleaned" folder above the dataset id?

youngdjn:
How about option #1 but paste back the dataset ID on the front afterward?

youngdjn (approved):
Awesome this looks great now! Thanks for all the iterating on this with me. You can go ahead and merge it.

PR #4, opened 2024-02-06
Sd metadata extraction

sdepaschalis:


sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+  # Calculate image frequency (images per second)
+  image_frequency <- 1 / time_diff_seconds
+
+  # Check if there are any non-finite image frequencies
+  if (any(!is.finite(image_frequency))) {
+    stop("Invalid image frequencies detected.")
+  }
+
+  # Calculate mean image frequency
+  mean_image_frequency <- mean(image_frequency, na.rm = TRUE)
```
youngdjn, line 44: Just before calculating the mean, it would be good to exclude outliers, like the time between the initial takeoff image and the first image of the transect. Presumably these would already have been removed during data curation, but it would be good to tolerate (and exclude) them if they exist. For this I recommend something like:

```R
image_frequency_lwr = quantile(image_frequency, 0.1)
image_frequency_upr = quantile(image_frequency, 0.9)
image_frequency_nooutliers = image_frequency[image_frequency > image_frequency_lwr & image_frequency < img_frequency_upr]
```

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+
+  # Get Xresolution and Yresolution from the EXIF data
+  resolution_x <- unique(exif$Xresolution)
+  resolution_y <- unique(exif$Yresolution)
+
+  # Get aspect ratio
+  aspect_ratio <- unique(exif$imagewidth / exif$imageheight)
```
youngdjn: I'd recommend using `Xresolution` and `Yresolution` attributes for aspect ratio also. Reasoning: if we need to adapte the code to a future drone or camera model that uses different names for the attributes, we only have to change two of them instead of four.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+
+  # Get image file format
+  image_file_format <- unique(exif$FileType)
+
+  # Check if resolution attributes are identical across all images
+  if (length(resolution_x) > 1 | length(resolution_y) > 1) {
+    # If not identical, compute average
+    avg_resolution_x <- mean(resolution_x)
+    avg_resolution_y <- mean(resolution_y)
```
youngdjn: You should calculate mean (or even better: mode) aspect ratio here too, in case there is more than one value across all the images in the dataset.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+  } else {
+    # If identical, use the single values
+    avg_resolution_x <- resolution_x
+    avg_resolution_y <- resolution_y
+  }
+
+  # Check if image file format is consistent
+  if (length(image_file_format) > 1) {
+    # If not identical, return a warning
+    warning("Image file format varies across images.")
```
youngdjn, line 75: Here, I would return the most common (mode) image file format, rather than the current approach which would return a vector of all of them.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+  # Check if there are any non-finite image frequencies
+  if (any(!is.finite(image_frequency))) {
+    stop("Invalid image frequencies detected.")
+  }
+
+  # Calculate mean image frequency
+  mean_image_frequency <- mean(image_frequency, na.rm = TRUE)
+
+  # Get Xresolution and Yresolution from the EXIF data
+  resolution_x <- unique(exif$Xresolution)
```
youngdjn: I like your approach of using `unique()` to determine if multiple resolutions exist. The only downside is that it drops information about the frequency of different resolutions (e.g. maybe there's only one outlier), and if we're in a case where we have multiple resolutions, we'd probably want to return the mode (most common) resolution. The current approach  takes the average of all the unique values, which is a little different. I would still use `unique()` and use the length of the result to determine if there are multiple resolutions, but I would do a separate operation to determine the mode, which would be the value to return. This applies to aspect ratio and file format too.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -0,0 +1,104 @@
...
+datadir = readLines(file.path("sandbox", "data-dirs", "steven-metadata-laptop.txt"))
+
+exif_file = exif_files[1]
+exif = prep_exif(exif_file)
+
+# Load necessary packages
+library(sf)
+
+# Define the function to extract metadata including resolution, aspect ratio, image file format, and mean image frequency
+extract_metadata = function(exif) {
```
youngdjn: Please name this function `extract_metadata_sd` so that everyone working on this project can have their own functions and not conflict by creating multiple functions with the same mane. Then we can merge them later. Also note that as requested in the general comments, the content of this function should be converted into a few separate sub-functions, and this function `extract_metadata_sd` should simply be a wrapper around them. For example, see `extract_metadata_dy` in R/imagery-metadata-extraction_dy.R.

youngdjn (changes requested):
Fantastic work! I'm very impressed. I really like how you include checking for unexpected values. I left a few inline comments that should be straightforward to address. Here are a few overall comments as well:

- Rather than putting everything in an overall `extract_metadata` function, I'd request that you make separate functions for `extract_image_frequency`, `extract_resolution_and_aspect_ratio`, and `extract_file_type`. Then you can wrap each of these functions in a single overall `extract_metadata_sd` function (see examples in R/imagery-metadata-extraction_dy.R).
- The new function `extract_resolution_and_aspect_ratio` will have to return multiple values, which should be combined as a list first, just as you do for the `extract_metadata` function. The other functions can return a single value.
- There is a file with a weird name: `R/R:imagery-metadata-extraction_stevendepaschalis.R`. You need to remove the `R:` part of the filename.

Thank you!

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -49,56 +28,73 @@ extract_metadata = function(exif) {
...
-    # If identical, use the single values
-    avg_resolution_x <- resolution_x
-    avg_resolution_y <- resolution_y
-  }
-
-  # Check if image file format is consistent
-  if (length(image_file_format) > 1) {
-    # If not identical, return a warning
-    warning("Image file format varies across images.")
-  }
```
youngdjn: I really liked this check you originally included. It would be great to add it back. Including this is not mutually exclusive with returning the mode image type.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -49,56 +28,73 @@ extract_metadata = function(exif) {
...
   time_diff_seconds <- time_diff_seconds[is.finite(time_diff_seconds)]
 
-  # Calculate image frequency (images per second)
-  image_frequency <- 1 / time_diff_seconds
+  # Calculate median and median absolute deviation (MAD)
+  median_time_diff <- median(time_diff_seconds)
+  mad_time_diff <- mad(time_diff_seconds)
+
+  # Set threshold as median + 3 * MAD
+  threshold <- median_time_diff + 3 * mad_time_diff
```
youngdjn: What's the logic for using 3*MAD? I'm nor familiar with `mad`, and I'd be more comfortable with using quantiles as in my previous example unless you want to lobby for `mad`.

youngdjn (changes requested):
Looks great! Just two remaining small comments to address and then it will be ready to merge.

sandbox/extract-imagery-metadata_stevendepaschalis.R
```diff
@@ -87,6 +86,12 @@ extract_metadata_sd <- function(exif) {
   # Extract image file type
   image_file_format <- extract_file_type(exif)
 
+  # Check if image file format is consistent
+  if (length(unique(image_file_format)) > 1) {
+    # If not identical, return a warning
+    warning("Image file format varies across images.")
```
youngdjn: This needs to go inside the function `extract_file_type`, before calculating the mode.

youngdjn (changes requested):
It's *really* close now -- just one small inline comment.

youngdjn (approved):
Looks great! Please go ahead and merge to main.

PR #6, opened 2024-03-04
Emp

empurvis:
adding image level metadata extraction file in the .R folder; updating the sandbox image level metadata extraction file to remove the AGL function

### DIFF
--- a/R/extract-imagery-metadata_emp.R
+++ b/R/extract-imagery-metadata_emp.R
@@ -1,0 +1,313 @@
+# This script defines image-level metadata extraction functions
+# Written by Emily Marie Purvis, March 4th 2024
+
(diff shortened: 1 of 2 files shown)


R/extract-imagery-metadata_emp.R
youngdjn, line 1: This is great. You just need to delete the lines in each section that are not inside the function. (e.g. lines 8-12, 26-34, and the corresponding lines for the other functions). It makes sense to have these non-function lines for code development (e.g. in sandbox) but not for the actual package function definitions.

youngdjn (commented):
Looks great, just see my note on R/extract-imagery-metadata_emp.R. Thanks!

PR #7, opened 2024-03-05
Emp

empurvis:
fixed the requests!

### DIFF
--- a/R/extract-imagery-metadata_emp.R
+++ b/R/extract-imagery-metadata_emp.R
@@ -5,8 +5,6 @@
 
 #### dataset ID ####
 
-dataset_id = exif$dataset_id
-
 extract_dataset_id = function (exif) {
 
   dataset_id = exif$dataset_id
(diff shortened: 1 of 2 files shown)


empurvis:
darn, I accidentally committed another uncompleted draft file for the visualizations. it shouldn't matter, I'll put in another pull request when I finish that up

youngdjn (changes requested):
Most of the way there, but there is still code outside of functions. Sorry I was unclear. I didn't mean just remove comments, but actually remove all the code except what's in the functions. It's redundant with what's in the functions anyway (it was just used to develop the functions, but is now fully encapsulated in the accompanying functions). When this code file gets sourced upon loading the `ofo` package, the only way to access the functionality will be by calling the functions. Standalone code lines are extraneous.

empurvis:
ah gotcha. should be fixed now.

youngdjn (approved):
Awesome, looks great -- go ahead and merge! Then pull from main so you get all the other updates since you branched.

PR #9, opened 2024-03-21
Emp

empurvis:
Completed visualization code. HTMLs are on box!


R/extract-imagery-metadata_emp.R
```diff
@@ -42,6 +97,21 @@ extract_lat_lon = function (exif) {
 
 #### rtk_fix (Format: True/False. Use EXIF RTKFlat) ####
 
+#' Returns rtk_fix of image
+#'
+#' Pulls the rtk fix to include in image-level metadata collation. Return TRUE if there is a RtkFlag value of 50, and FALSE otherwise, i.e. other RtkFlag value, or none. IMPORTANT NOTE!!!!!! THIS FUNCTION ASSUMES DJI DRONES WERE USED. OTHER DRONE MAKES WILL REQUIRE AN UPDATED FUNCTION.
+#'
+#' @param rtk_fix if a column for "RtkFlag" is in the exif data, rtk_fix is true if the column value is 50 and false for any other value. If there isn't an "RtkFlag" column, rtk_fix is false.
```
youngdjn: `rtk_fix` is not a parameter of the function, just a variable used internally that happens to be returned, so we should remove this doc line

R/extract-imagery-metadata_emp.R
```diff
@@ -15,6 +32,23 @@ extract_dataset_id = function (exif) {
 
 #### datatime_local (Format: YYYYMMDD HHMMSS (local time zone, 24 hr)) ####
 
+#' Extract local date and time of image collection
+#'
+#' Pulls the local date and time of image collection (Format: YYYYMMDD HHMMSS) to include in image-level metadata collation
+#'
+#' @param exif$DateTimeOriginal_nocolon local date and time of image collection
+#'
+#' @param datatime_local local date and time of image collection
```
youngdjn: The only parameter of this function is `exif` so the definitions of the others should be deleted. The others are just internal variables.

R/extract-imagery-metadata_emp.R
```diff
@@ -53,7 +123,26 @@ extract_rtk_fix = function(exif) {
...
+#'
+#' @param accuracy_x returns the RTK standard longitude deviation in meters, and returns NA if the RtkStdLon column does not exist
+#'
+#' @param accuracy_y returns the RTK standard latitude deviation in meters, and returns NA if the RtkStdLat column does not exist
+#'
+#' @param accuracy returns a data.frame of the accuracy_x, accuracy_x, and accuracy_z values for each image
+#'
+#' @param exif the exif metadata file
+#'
+#' @return accuracy of latitude, longitude, and height of each image. Standard deviation. Units: meters.
```
youngdjn: No longer returning height

R/extract-imagery-metadata_emp.R
youngdjn, line 1: You're going to find this really annoying. You were very thorough in documenting the functions, but I'm afraid you were actually over-thorough. Many of your `@param` definitions are just for variables used internally in the function, so they don't need to be (and shouldn't be) defined in the documentation, because a user would have no knowledge of them or ability to interact with them. The only params that should be defined are those that are passed to the function (in most cases, just `exif`). Sorry but I'm afraid we need to remove the other definitions! The `@return` (and other attribute) documentation is good to keep though!

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
 
+exif4$received_image_path <- extract_received_image_path(exif4)
 
-exif_file = exif_files[1]
+exif5$received_image_path <- extract_received_image_path(exif5)
 
-metadata1 <- ofo::extract_metadata_emp(exif_file)
+exif6$received_image_path <- extract_received_image_path(exif6)
 
+# now let's merge the exif files with the metadata files
```
youngdjn: It's unclear to me why you can't just use the metadata objects. Why do they have to be joined to the exif objects? If there are additional attributes you need out of the exif objects, I would write more `extract_*` functions and add them to `extract_metadata_emp` to get them into the `metadata` objects.

sandbox/visualizations_emp.R
youngdjn, line 1: You take the approach of prepping all 6 example exif (metadata) objects in advance. While this is not *wrong* (and in fact very thorough), I would take a slightly simpler approach of just loading one at the top and running the rest of the script through with it, then changing it at the top to a second one, and running the script through again. This approach would lend itself well to turning the map-making logic into a function that you could call and pass it whatever `exif` or `metadata` object you want. Then you could loop over all `metadata` objects, calling the function for each one. The purpose of the "sandbox" scripts is to develop functionality that will ultimately be turned into functions defined in the R folder of this repository, which will make them accessible to anyone who loads the `ofo` package. So there's no "right" or "wrong" way of writing code in `sandbox`, but it may make your life easier to consider where the code will ultimately end up (as a function that can be run on one `metadata` object).

sandbox/visualizations_emp.R
youngdjn: OK after reading on, I think I understand now! You need some attributes, like date/time, in a format that is specific to the leaflet mapping, and not needed as metadata otherwise. OK I think I'm back on board! One idea, (not sure if it's good) is instead of calling `extract_metadata_emp` to create a comprehensive `metadata` object, what about just creating one with the specific attributes you need, by calling the `extract_*` functions directly? If you just need a few attributes, this might be more effeicient that extracting *all* the metadata that we could possibly report. I'd advocate for switching to this because ultimately, I think our overall `extract_metadata` function will be pretty slow because of functionality not needed here (e.g., one step will be to download a DEM to see how well the drone flight tracked the terrain).

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+totalexif5$datetime <- ymd_hms(totalexif5$DateTimeOriginal)
 
-## compiled image-level metadata. after loading each metadata file, add a row index ("X") identical to the row index in the original data files-- this will allow us to easily merge the two types of files in the next step.
+totalexif6$datetime <- ymd_hms(totalexif6$DateTimeOriginal)
 
-exif_20220630_0041_metadata <- read.csv ("extracted-metadata\\image-level-metadata\\image-metadata_20220630-0041.csv")
+# creating "duration in seconds from beginning of image to collection to time of [i] image" column. the number in brackets is the index position of the image that was collected first.
 
-exif_20220630_0041_metadata$X <- 1:nrow(exif_20220630_0041_metadata)
+totalexif1$duration <- as.duration(totalexif1$datetime[1] %--% totalexif1$datetime)
```
youngdjn: Unclear why you can't just use `-` instead of %--%, but I'm probably missing something!

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+    "Collection date and time:", totalexif1$DateTimeOriginal, "<br>",
+    "Drone altitude above sea level:", totalexif1$altitude_asl, " meters",        "<br>"),
+    radius = 2,
+    weight = 1,
+    stroke = TRUE,
+    fillOpacity = 1,
+    color = "#008080"
+    ) %>%
+  # add secret invisible circles at each image location with a bigger radius so they're actually clickable
+  addCircleMarkers(
```
youngdjn: Nice!!

youngdjn: I noticed they got easier to click!

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+    zoom = 16
+    ) %>%
+  # add circles at each image location
+  addCircleMarkers(
+    ~lon,
+    ~lat,
+    popup = paste(
+    "Image filepath:", totalexif1$received_image_path, "<br>",
+    "Collection date and time:", totalexif1$DateTimeOriginal, "<br>",
+    "Drone altitude above sea level:", totalexif1$altitude_asl, " meters",        "<br>"),
```
youngdjn: If these are not the circles that we want to be clickable, can we drop the popup bit, given there are now the larger clickable circles?

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+exif1_MAP <-
+  # create leaflet function using data specific to this map
+  leaflet(totalexif1) %>%
+  # use the jawg.terrain basemap
+  addTiles(
+    'https://tile.jawg.io/jawg-terrain/{z}/{x}/{y}{r}.png?access-token=cN5iUzB2bL1QMdeVwMJlflwXQweOoNTDOadGfoCbPf1rRM9FGPb4i1m4f2k5MaJP') %>%
+  # set the center of the map to be the center ish of the flight footprint
+  setView(
+    lng=(totalexif1$lon[888]),
+    lat=(totalexif1$lat[888]),
```
youngdjn: I'd prefer for us to instead calculate a spatial centroid as part of `extract_metadata` (i.e. a new function `extract_centroid`). This will be a useful attribute for other purposes too I think.

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+    zoom = 16
+    ) %>%
+  # add circles at each image location
+  addCircleMarkers(
```
youngdjn: If this layer is not displaying AGL or duration, what is its purpose. I don't understand leaflet well enough to understand why this is necessary.

sandbox/visualizations_emp.R
```diff
@@ -11,176 +11,368 @@ library(leaflet.providers)
...
+    stroke = FALSE,
+    fillOpacity = 0
+  ) %>%
+  # add an interactive layer for ASL
+  addCircleMarkers(
+    ~lon,
+    ~lat,
+    popup = paste(
+      "Image filepath:", totalexif1$received_image_path, "<br>",
+      "Drone altitude above sea level:", totalexif1$altitude_asl, " meters",        "<br>"),
```
youngdjn: Can we drop the popup here since it's already included in the larger invisible clickable circles? Same for the collection time circles.

sandbox/visualizations_emp.R
youngdjn: Therefore I only commented on the first leaflet map being created. Interpret my comments as applying to all iterations. Now that there are a few changes to make (to each one), it's more clear how making this into a function that you just call 6 times (each with a different EXIF) could save some headaches.

youngdjn (changes requested):
Awesome! Just a few inline comments that should be pretty easy to address -- with one exception of turning the mapmaking code into a function -- but even that should be pretty straightforward I think. Thanks for all your work on this!

R/extract-imagery-metadata_emp.R
empurvis: yes you're totally right! I was struggling to get the metadata extraction function to work and thought I'd randomly change the documentation to add all this other stuff in to see if that was the problem, which it obviously wasn't. I'll take it back out and update the documentation in the man folder

sandbox/visualizations_emp.R
empurvis: Yes, precisely. Leaflet doesn't like the date/time format from the metadata extraction function, and I couldn't manipulate that value to get what I needed, so I resorted to working with the original exif "DateTimeOriginal" value. I don't think we need to add anything new to the extract_metadata_emp set of functions-- on the contrary I think might be confusing to have multiple date/time formats.

sandbox/visualizations_emp.R
empurvis: I checked the google doc and Michelle is working on this! I can write in a really quick bonus calculation in this visualization code that doesn't rely on her function, as it's not in the main branch yet and I have no clue what the status is. I already have to do bonus calculations with date/time data so it should be no big deal. I can take the mean of the image lat and lon columns from the set of imagery metadata extraction functions.

sandbox/visualizations_emp.R
empurvis: I can definitely change the code so that I'm calling individual extract_* functions rather than the whole extract_metadata_emp function! I think I only need 5 or so of those extracted values. I will still have to merge them together and include the date/time calculations that rely on the original exif.

sandbox/visualizations_emp.R
empurvis: No clue, I just followed instructions from the lubridate package documentation to the letter. The "-" doesn't do the same thing.

empurvis: I just left it as is, as it seems like no big deal, but let me know if you want me to dig into this more and try to change it

sandbox/visualizations_emp.R
empurvis: these are just the base circles that exist when both AGL and duration layers are unclicked. they're the same size and in the same place as the other circles. it just means the contents of the map don't entirely pop out of existence when both interactive layers are turned off. I think this is a useful/intuitive feature but I can definitely get rid of these if you want.

sandbox/visualizations_emp.R
empurvis: done!

empurvis:
edits made, new commits should address all the issues

sandbox/visualizations_emp.R
youngdjn: Got it -- makes sense, thanks.

youngdjn (approved):
Looks great! This is good to merge into main!

PR #12, opened 2024-04-10
Mg metadata extraction

MishGarcia:


sandbox/extract-imagery-metadata_mg.R
```diff
@@ -0,0 +1,181 @@
...
+GPSLocation = exif$GPSPosition
+
+#splitting the list (was this step really necessary for my dataset? I'm not really sure this actually helped?)
+coordinates <- strsplit(GPSLocation, split = ";")
+
+# Separating the GPS from the whole exif file
+almostdone <- strsplit(unlist(coordinates), split = ",")
+
+#Creating a seperate dataframe with just the coordinates to call upon later
+isolatedcoordinates <- as.data.frame(do.call(rbind, almostdone))
```
youngdjn: It looks like this code is all obsolete because nothing below depends on it, so it can be removed.

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -0,0 +1,181 @@
...
+
+install.packages("tidyverse")
+library(tidyverse)
+
+df2 = separate(exif,
+        col = GPSPosition,
+        into = c("lat", "long"),
+        sep = " " )
+#Here this worked, instead of doing by number of characters, did it by seperating at
+#the space in between the two (since the number of characters differed from row to row).
```
youngdjn: Nice work getting lat/lon this way! However, because `exif` is a spatial object, it is slightly preferable to use the following approach:

```
coords = sf::st_coordinates(exif)
lon = coords[, 1]
lat = coords[, 2]
```
This will make sure your code (at least this part of it) is robust to future drone models that might have EXIF columns named differently. You can trust that whatever the right column names were, they were already used for obtaining the lat/lon that is embedded in the `exif` spatial object.

youngdjn: Even better, you would compute the centroid as a geospatial `sf` object, which you could do by using some functions in the `sf` package applied to the `exif` object, without ever having to extract coordinates. See if you can find any suggestions for this workflow online, and if not I can hep with it.

youngdjn: Then you could pass this centroid object directly to the solarnoon function (it accepts `sf` objects).

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -0,0 +1,181 @@
...
+#character.
+
+xcoordinate = mean(df2$lat, na.rm=TRUE)
+ycoordinate = mean(df2$long, na.rm= TRUE)
+
+#finding the mean of the lat and long, since this is all one mission,
+#it should give us the centroid lat/long, at least from what I researched.
+
+print(ycoordinate)
+print(xcoordinate)
```
youngdjn: Once you have computed the centroid as a `sf` object, you an simply return that object, rather than two objects containing the text of the coordinates.

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -0,0 +1,181 @@
...
+
+
+
+#START CODE FOR SOLAR NOON CALCULATIONS
+
+#found the package suntools which will calculate the solar noon, this way we do not have to do
+#any crazy calculations, it is all included in the function and only needs lat/long and the time.
+
+install.packages("suntools")
+library("suntools")
```
youngdjn: Nice work finding this package! It looks perfect for this task.

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -0,0 +1,181 @@
...
+date <- datetimedf[1, 143]$date
+date
+#now using the newly seperated date column and using the first row to get the date to be used in the function below
+
+#used the centroid calculated in the above code chunk for the lat/long
+solarnoon <- solarnoon(
+  matrix(c(xcoordinate, ycoordinate), nrow = 1),
+  as.POSIXct(date),
+  POSIXct.out=TRUE
+)
```
youngdjn: When I ran it, this function said solarnoon was 2 am :P . I tested what happened when I swapped the x and y coords, and it gave a reasonable result. Apparently it expects the reverse order. I didn't see this in the documentation anywhere though!

sandbox/extract-imagery-metadata_mg.R
youngdjn: Please check for any other obsolete code too.

youngdjn (commented):
Great work! Thanks! I just have a few inline suggestions below. Once you address them, the next step will be turning these lines of code into functions that take `exif` as an argument and return their respective calculated values (see examples above in this code file). Let me know if you have any questions.

One general comment is to please put all package load calls at the top of the script. Also, please don't include `install.packages` calls. These just have to be run once per user, so they shouldn't be part of the main workflow.

youngdjn (changes requested):
I meant to save this as a "request changes" review but accidentally saved it as a "comment review".

sandbox/extract-imagery-metadata_mg.R
mikoontz: Agree with @youngdjn here about calculating the centroid directly, then passing that to the suntools::solarnoon() calculation. It does look like it accepts {sf} objects (https://github.com/adokter/suntools/blob/374fff7b2f595fabb8bf75d928c63b6e2f099e98/R/solarnoon.R#L16-L42). Nice find on this package!

Note for posterity that latitude is actually the 2nd column in the result of `sf::st_coordinates(exif)`, not the first. So Derek's code would be: 
```
coords = sf::st_coordinates(exif)
lon = coords[, 1]
lat = coords[, 2]
```

youngdjn: Oops I always do that! Updating my comment to minimize future confusion.

sandbox/extract-imagery-metadata_mg.R
mikoontz: I think that `suntools::solarnoon()` does require x, then y, but I think the original extraction of those coordinates using `tidyr::separate()` earlier in the code assigned latitude to `xcoordinate` and longitude to `ycoordinate` so they were passed backwards here in L166 even though the labeling is correct. Another good reason to take Derek's suggestion and calculate the centroid of all of the lon/lat and then pass the centroid as an {sf} object. Something like:

```
exif_centroid = sf::st_union(exif) |> sf::st_centroid()
```

`sf::st_union()` takes all the individual points and collapses them into a single MULTIPOINT. Then the `sf::st_centroid()` function takes the mean of all the latitudes and the mean of all the longitudes (as you discovered) to calculate the centroid.

Does the `suntools::solarnoon()` function expect that the coordinate reference system (CRS) of the {sf} object is something specific? If so, you might want to use `sf::st_transform()` to transform the centroid to the expected CRS. But maybe the `suntools()` function does that for you!

If I had a nickel for every time I switched lon/lat and x/y...

You might also consider assigning the result of the `suntools::solarnoon()` function call to something other than `solarnoon` since you are overwriting the `suntools::solarnoon()` function with the calculation! Like `exif_solarnoon` or `flight_solarnoon` (whatever the unit is represented by the `exif` object.

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -84,58 +86,28 @@ image_count
...
-        sep = " " )
-#Here this worked, instead of doing by number of characters, did it by seperating at
-#the space in between the two (since the number of characters differed from row to row).
+#Getting the coordinates listed in the exif file and for it to seperate into long/lat
 
-print(df2)
-#printed out the new df that includes the "lat" and "long"
+coordinate = sf::st_coordinates(exif)
+lon = coords[, 1]
+lat = coords[, 2]
```
youngdjn: You no longer need `lon` or `lat`, and also the `coords` object is not defined (only `coordinate` is defined) so these lines generate errors. You can delete these lines.

sandbox/extract-imagery-metadata_mg.R
youngdjn, line 1: Please delete the obsolete lines 
```
datetime <- df2[1, "capture_datetime"]
datetime
```

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -161,21 +132,27 @@ date <- datetimedf[1, 143]$date
...
+#used the general coordinates and input into this function
+
+sncalc <- solarnoon(
+  matrix(c(coordinate), nrow = 1),
   as.POSIXct(date),
   POSIXct.out=TRUE
 )
 
-solarnoon
+sncalc
```
youngdjn, line 50: Hooray, this appears to be working great!

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -161,21 +132,27 @@ date <- datetimedf[1, 143]$date
...
 #END CODE FOR SOLARNOON CALCULATION
 
 
+#START CODE FOR EARLIEST/LATEST TIME
 
+earliest_time_local <- min(exif$capture_datetime)
+earliest_time_local
 
+latest_time_local<- max(exif$capture_datetime)
+latest_time_local
```
youngdjn: This looks good but the output needs to be in the format specified in the protocol:

`YYYYMMDD HHMMSS (local time zone, 24 hr)`

Currently it is a POSIXct datetime object that prints as `"2022-06-30 14:32:47 UTC"`, but it needs to be turned into a pure string (character vector) formatted as above.

sandbox/extract-imagery-metadata_mg.R
```diff
@@ -161,21 +132,27 @@ date <- datetimedf[1, 143]$date
 date
 #now using the newly seperated date column and using the first row to get the date to be used in the function below
 
-#used the centroid calculated in the above code chunk for the lat/long
-solarnoon <- solarnoon(
-  matrix(c(xcoordinate, ycoordinate), nrow = 1),
+#used the general coordinates and input into this function
+
+sncalc <- solarnoon(
+  matrix(c(coordinate), nrow = 1),
```
youngdjn: Please keep formatting consistent (first argument on same line as function name, closing parenthesis on same line as last argument).

youngdjn (changes requested):
All the logic here is great! Everything is working (except a few obsolete lines that generate errors), just a few revisions requested below.

When revising, please also delete the example code at the top (copied in from my template file). You can always refer to it in the original template file if needed.

After addressing these comments, the next step (either before or after the next review) is to move the extraction code for these two metadata attributes into their own functions.

R/R:imagery-metadata-extraction_mg.R
youngdjn, line 1: I'm not clear on why this file got added. It appears to be a duplicate of my example file. It also has an odd `R:` in the filename.

sandbox/extract-imagery-metadata_mg.R
youngdjn: These lines were not removed

sandbox/extract-imagery-metadata_mg.R
youngdjn: This change was not made

youngdjn (commented):
Hi @MishGarcia , your newest commit does not appear to have any changes in it other than an additional blank line. Did you save your scripts before committing and pushing? Your commit message included "ready for review" so I assumed you wanted me to review it. Let me know if you want to troubleshoot.

youngdjn (approved):
Nice work! This is just missing a `library(sf)` and a `exif = prep_exif(exif_file)`, but I can take care of that after it is merged. I'm ready to go ahead and use this code, so I'm going to merge it now and make the final tweaks to integrate it into our codebase (including turning these steps into standalone functions that can be called). Thanks for all your work on this!

PR #16, opened 2024-04-17
Sd metadata extraction

sdepaschalis:


imagery-metadata-extraction_sd2.R
youngdjn, line 1: I don't think this will work quite right, for a couple reasons:
- For smart-oblique missions, it's possible the straight-down (nadir) shots will be taken with a camera pitch that is close to nadir but not exactly nadir. I think anything that is +- 2 degrees of nadir should be excluded.
- Nadir is represented in the EXIF as `-90`, not as `0`, so you will probably have to add 90. Please confirm this is the case across all 6 of the example EXIF files (which intentionally come from a range of drones and mission types).
- The smart-oblique shots include a forward and backward shot. After adding 90, these would be symmetrical around zero (e.g. -30 and 30). So you can't just average them, or it will average out to zero. It should work to compute the absolute value and then average them.
- You will need to think about how to deal with missions that were intentionally shot with a nadir camera (only). You don't want to filter out all those zeros, or you'll have no data left. I think you will need to detect if it is smart oblique or not (based on there being 3 different pitches with relatively even frequency) and process accordingly.

imagery-metadata-extraction_sd1.R
youngdjn, line 1: Logic looks good!

youngdjn (changes requested):
Thanks! I left comments on each file below, and here are a few general comments:
- These two scripts are in the top-level folder of the repo. Please move them to `sandbox`, and ideally combine into one file.
- Once you have addressed the comments below, the next step is to put this logic inside a functions (one for the datetimes, and one for the pitch) that accepts `exif` as an argument and returns the calculated/extracted values. For an example, see the template file referenced from the protocol document for this task.

Thank you!

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -17,20 +17,28 @@ process_pitch_values <- function(pitch_values) {
...
+    # Check if the range between 0.1 and 0.9 quantiles is at least 10 degrees
+    pitch_range = quantiles[3] - quantiles[1]
+    if (pitch_range >= 10) {
+      # Check if the absolute values of 0.1 and 0.9 quantiles are similar
+      if (abs(quantiles[1]) - abs(quantiles[3]) < 5) {
+        # Use 0.1 or 0.9 quantile directly as the mission pitch
+        processed_pitch = quantiles[1]
+      } else {
+        # Compute the average of absolute pitch values for smart oblique missions
+        processed_pitch = mean(abs(filtered_pitch_values), na.rm = TRUE)
```
youngdjn: Nice, this should work since `filtered_pitch_values` should exclude near-nadir (+-2 deg), but it would be susceptible to including a few low-pitch (e.g. 5 deg) images, for example if they were taken manually. Therefore I'd change this slightly to:

`processed_pitch = mean(abs(c(quantiles[1], quantiles[3])), na.rm = TRUE)`

Then you wouldn't even need to compute `filtered_pitch_values`. You could instead compute and use a something like `adjusted_pitch_values`, defined as `pitch_values + 90`.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -17,20 +17,28 @@ process_pitch_values <- function(pitch_values) {
...
+      # Use the regular mean for non-smart oblique missions
+      processed_pitch = mean(filtered_pitch_values, na.rm = TRUE)
+    }
   } else {
-    # Use the regular mean for non-smart oblique missions
-    processed_pitch = mean(filtered_pitch_values, na.rm = TRUE)
+    processed_pitch = NA  # Set processed_pitch to NA if quantiles are not valid
   }
 
   return(processed_pitch)
```
youngdjn: Pleas make sure that this return value uses 0 to represent nadir (down). That is, you should use `adjusted_pitch_values` or similar, as I suggest above.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -17,20 +17,28 @@ process_pitch_values <- function(pitch_values) {
   # Exclude pitch values within ±2 degrees of Nadir (-90 degrees)
   filtered_pitch_values = pitch_values[abs(pitch_values + 90) > 2]
```
youngdjn: You don't actually have to filter pitch values for the logic below to work (after implementing one small change I suggest below). However, I think you should use a new variable `adjusted_pitch_values = pitch_values + 90` for all the logic below, so that 0 represents down (nadir).

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -32,13 +26,40 @@ process_pitch_values <- function(pitch_values) {
...
-for (exif_file in exif_files) {
-  exif = prep_exif(exif_file)
+# Define a function to extract datetime and pitch values from EXIF data
+extract_metadata <- function(exif) {
+  # Convert DateTimeOriginal to datetime objects
+  exif$DateTimeOriginal <- ymd_hms(gsub(":", " ", exif$DateTimeOriginal), tz = "UTC")
+
+  # Calculate start and end datetime_local
+  start_datetime_local <- min(exif$DateTimeOriginal)
+  end_datetime_local <- max(exif$DateTimeOriginal)
```
youngdjn: These lines, plus the two additional datetime formatting lines below, should be moved to their own function, the same way `process_pitch_values` is its own function.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -32,13 +26,40 @@ process_pitch_values <- function(pitch_values) {
...
+
+  # Process pitch values
   camera_pitch_values = exif$CameraPitch
   camera_pitch_derived = process_pitch_values(camera_pitch_values)
```
youngdjn: Please modify the function so that it takes `exif` as an argument directly, rather than having to pull the `CameraPitch` attribute first. See my example exif attribute extraction function `extract_flight_speed` [here](https://github.com/open-forest-observatory/ofo-r/blob/4f22ca71bf72078b9307403ad24caf1510f30120/R/imagery-metadata-extraction_dy.R) as an example.

youngdjn (changes requested):
Nice, more great progress! The pitch extraction logic is almost spot on. A few follow-up comments below.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -6,66 +6,62 @@ library(lubridate)
...
-        # Compute the average of absolute pitch values for smart oblique missions
-        processed_pitch = mean(abs(filtered_pitch_values), na.rm = TRUE)
-      }
+      # Use 0.1 or 0.9 quantile directly as the mission pitch
+      processed_pitch <- quantiles[1]
     } else {
-      # Use the regular mean for non-smart oblique missions
-      processed_pitch = mean(filtered_pitch_values, na.rm = TRUE)
+      # Compute the average of absolute pitch values for non-smart oblique missions
+      processed_pitch <- mean(abs(c(quantiles[1], quantiles[3])), na.rm = TRUE)
```
youngdjn: This should just be `median(adjusted_pitch_values)`, since the quantiles are not relevant when the mission is not smart oblique.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -6,66 +6,62 @@ library(lubridate)
...
     if (pitch_range >= 10) {
-      # Check if the absolute values of 0.1 and 0.9 quantiles are similar
-      if (abs(quantiles[1]) - abs(quantiles[3]) < 5) {
-        # Use 0.1 or 0.9 quantile directly as the mission pitch
-        processed_pitch = quantiles[1]
-      } else {
-        # Compute the average of absolute pitch values for smart oblique missions
-        processed_pitch = mean(abs(filtered_pitch_values), na.rm = TRUE)
-      }
+      # Use 0.1 or 0.9 quantile directly as the mission pitch
```
youngdjn: Please add a comment saying we infer the mission is a smart-oblique mission, and one below in the `else` section that says we infer it is not smart-oblique.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -6,66 +6,62 @@ library(lubridate)
...
+    processed_pitch <- NA  # Set processed_pitch to NA if quantiles are not valid
   }
 
   return(processed_pitch)
 }
 
-# Define a function to extract datetime and pitch values from EXIF data
-extract_metadata <- function(exif) {
+# Define a function to process datetime values
+process_datetime_values <- function(exif) {
```
youngdjn: For consistency with the other extraction functions, please name this function: `extract_start_end_datetime`.

sandbox/imagery-metadata-extraction_sd2.R
```diff
@@ -6,66 +6,62 @@ library(lubridate)
...
-# Get a list of the files containing the test EXIF data (one file per image dataset). These files
-# have already been created and saved into the project data folder.
-exif_files = list.files(file.path(datadir, "exif-examples"), pattern = "^exif.+\\.csv$", full.names = TRUE)
+datadir <- readLines(file.path("sandbox", "data-dirs", "steven-metadata-laptop.txt"))
 
 # Define a function to process pitch values
-process_pitch_values <- function(pitch_values) {
-  # Exclude pitch values within ±2 degrees of Nadir (-90 degrees)
-  filtered_pitch_values = pitch_values[abs(pitch_values + 90) > 2]
+process_pitch_values <- function(exif) {
```
youngdjn: For consistency with the other extraction functions, please name this function: `extract_pitch`.

youngdjn (changes requested):
Thanks, almost there! Please see specifics below, and also work on the general organization we just discussed (R vs sandbox) and the file names.

youngdjn (approved):
Looks good! Thanks for cleaning it up. This is ready to merge -- please go ahead and merge it to main. Thanks again for all your work on this...now neatly wrapped up.

PR #18, opened 2024-04-29
EMP imagery metadata extraction functions

empurvis:
completed dataset level metadata extraction. new files in both the R and sandbox folders plus accompanying documentation in the man folder. re-named the image-level "extract_dataset_id" function; made minor changes to the relevant files in the R and sandbox folders plus the documentation.


sandbox/extract-dataset-metadata_emp.R
```diff
@@ -6,6 +6,7 @@
 
 library(stringr)
 library(tidyverse)
+library(sf)
 library(devtools)
 
 # Set working directory
```
youngdjn: This should not be necessary if you use `datadir` as it was set up in the example (and how you are mostly using it!).

sandbox/extract-dataset-metadata_emp.R
```diff
@@ -88,4 +89,324 @@ image_count = extract_image_count(exif)
...
+
+percent_images_rtk = (sum(exif$rtk_fix == TRUE) / nrow(exif)) * 100
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_percent_images_rtk = function (exif) {
+
+  exif$rtk_fix = extract_rtk_fix(exif)
```
youngdjn: You don't have to save this back to the `exif` data frame as a new column, you can just create a new variable `rtk_fix`, since this function will not actually affect the data frame outside the function.

sandbox/extract-dataset-metadata_emp.R
```diff
@@ -88,4 +89,324 @@ image_count = extract_image_count(exif)
...
+
+white_balance_mode_derived = Mode(exif$white_balance)
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_white_balance_mode_derived = function (exif) {
+
+  exif$white_balance = extract_white_balance(exif)
```
youngdjn: Also here, this does not have to be saved back to the exif dataframe.

sandbox/extract-dataset-metadata_emp.R
```diff
@@ -88,4 +89,324 @@ image_count = extract_image_count(exif)
...
+
+# The proportion of images with a white balance setting that matches the modal white balance setting in the dataset
+
+# BEGIN FUNCTION CODE
+
+exif$white_balance = extract_white_balance(exif)
+
+white_balance_mode_derived = extract_white_balance_mode_derived(exif)
+
+white_balance_mode_prop_derived = (sum(exif$white_balance == white_balance_mode_derived) / nrow(exif))
```
youngdjn: I would merge the additional logic here into the function `extract_white_balance_mode_derived` to reduce redundancy, and have that function (which you should rename) return a list of two named values.

sandbox/extract-dataset-metadata_emp.R
```diff
@@ -88,4 +89,324 @@ image_count = extract_image_count(exif)
...
+
+image_density = (nrow(exif))/(units::set_units(sf::st_area(create_mission_polygon(exif, image_merge_distance)), "hectare"))
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_image_density = function (exif, image_merge_distance) {
+
+  image_density = (nrow(exif))/(units::set_units(sf::st_area(create_mission_polygon(exif, image_merge_distance)), "hectare"))
```
youngdjn: Nice, I love how you're using units here!

sandbox/extract-dataset-metadata_emp.R
```diff
@@ -88,4 +89,324 @@ image_count = extract_image_count(exif)
...
+
+# BEGIN FUNCTION CODE
+
+area_ha = units::set_units(sf::st_area(create_mission_polygon(exif, image_merge_distance)), "hectare")
+
+# END FUNCTION CODE
+
+# turn code into a function
+
+extract_area_ha = function (exif, image_merge_distance) {
```
youngdjn: I would merge this with the density function to reduce redundancy, and have the new function return a named list of two values.

sandbox/extract-imagery-metadata_emp.R
youngdjn, line 1: In `extract_received_image_path`, you need to add `stringr::` preceding the `fixed` function, and you need to rename `dataset_id` to `dataset_id_image_level`. You also need to add this as an argument of the function `extract_received_image_path` so that it is accessible inside the function. Alternatively (maybe preferred), inside the function you can just access the dataset_id via `exif$dataset_id`.

R/extract-dataset-metadata_emp.R
youngdjn, line 1: Please rename to `dataset-metadata-extraction_emp.R`

R/extract-imagery-metadata_emp.R
youngdjn, line 1: Please rename to `imagery-metadata-extraction_emp.R`

youngdjn (changes requested):
Awesome work! With one exception it all runs super smoothly and generates the expected values. Thanks for all this work. My comments are all minor and mostly around streamlining. I made my comments mostly on the `sandbox/` files, but please make sure that any edits also get applied to the `R/` files. BTW at this stage, it's OK if you remove the function logic and function definitions from the sandbox scripts entirely (so you don't have to update it in multiple places). I intended sandbox mostly for your own use while developing the functions. You should at least keep the part of the sandbox scripts that call the functions (defined in `R/`) so that you and I have an easy way to test them.

sandbox/extract-dataset-metadata_emp.R
empurvis: theoretically yes.....but in practice this doesn't work for me (for mysterious unknown reasons). happy to remove these library calls though!

empurvis: I feel like we talked about this at some point and maybe you were having problems with packages external to ofo too?

sandbox/extract-imagery-metadata_emp.R
empurvis: hmm, all these things (except for the second stringr:: I forgot) are already done in my current versions which should be the versions on github already. I'll commit again and see if that fixes them.

sandbox/extract-dataset-metadata_emp.R
youngdjn: Oops I meant not the package loads, but the line after them that sets the working directory.

youngdjn (approved):
I think this is now good to go! Please go ahead and merge. Thanks again for all this great work.

I just noticed tiny issue I wanted to make you aware of, but which can easily be fixed once merged. When I run `devtools::load_all()`, I get the following warning:

```
Warning message:
Objects listed as exports, but not present in namespace:
• extract_area_ha
• extract_image_density
• extract_white_balance_mode_derived 
```

I guess these old, removed functions still show up in `NAMESPACE`. I'm not sure if this requires manual removal, or if running `devtools::document()` would remove them automatically. I'll check this once I combine everyone's metadata functions.

PR #31, opened 2024-09-03
Add Hyyppä and Muhojoki feature-based registration algorithm

russelldj:
This algorithm is proposed in Hyyppä and Muhojoki et al (2021), "Efficient coarse registration method using translation- and rotation-invariant local descriptors towards fully automated forest inventory", ISPRS Open Journal of Photogrammetry and Remote Sensing.

This implementation is very closely based on a MATLAB version released by the authors [here](https://gitlab.com/fgi_nls/public/2d-registration/-/tree/main?ref_type=heads).

### DIFF
--- a/R/tree-map-alignment.R
+++ b/R/tree-map-alignment.R
@@ -569,7 +569,7 @@ compute_feature_descriptors = function(xy_mat, R_local){
     feat_desc_mat[, i_quadrant+4] = angle_quad_i_norm;
   }
 
-  return(feature_desc_mat)
+  return(feature_desc_mat, char_dirs)
 }
 
 # Implementation of the following MATLAB function
--- a/R/tree-map-alignment.R
+++ b/R/tree-map-alignment.R
@@ -472,6 +472,67 @@ find_best_shift_icp = function(pred, obs) {
 }
 
 # Implementation of the following MATLAB function
+# https://gitlab.com/fgi_nls/public/2d-registration/-/blame/main/fit_euclidean_transformation.m?ref_type=heads#L231
+compute_feature_descriptors = function(xy_mat, R_local){
+  xy_mat = as.matrix(xy_mat)
+  # Convert the pairwise distances between each row
+  pdist = dist(xy_mat, method = "euclidean")
+  # Convert from a dist object into a matrix
+  pdist = as.matrix(pdist)
+  # Set the diagnal to the max value so it will never be the minimum
+  diag(pdist) <- max(pdist)
+  # Finding the closest neighboring object for each object
+  # Invert the sign so we can use max.col
+  indices_of_closest_points = max.col(-pdist)
+
+  # Coordinates of the closest neighboring object for each of the objects
+  closest_points = xy_mat[indices_of_closest_points, ]
+  # For each object, compute the normalized characteristic directions that
+  # are locally used as the 1st axis direction.
+
+  char_dirs = closest_points - xy_mat
+  distances = sqrt(char_dirs[,1]^ 2 + char_dirs[,2]^2)
+  char_dirs = char_dirs / distances
+
+  # Directions perpendicular to the characteristic directions. These are
+  # used as the local 2nd axis direction.
+  char_dirs_perp = cbind(-1 * char_dirs[,2], char_dirs[,1])
+
+
+  # For each object, we transform the coordinates of the other objects
+  # into the local coordinate frame of the given object. We denote the
+  # coordinate along the 1st axis direction by v and the coordinate along
+  # the 2nd axis direction by w. In the matrices below, the element at row
+  # i and column j means the coordinate of the jth object in the local
+  # coordinate frame of the ith object.
+
+  # Define intermediate variables
+  x_values = xy_mat[,1]
+  y_values = xy_mat[,2]
+
+  n_points = nrow(xy_mat)
+  x_tiled = matrix(rep(t(x_values), n_points), ncol=n_points)
+  y_tiled = matrix(rep(t(y_values), n_points), ncol=n_points)
+  x_minus_x_transposed = x_tiled - t(x_tiled)
+  y_minus_y_transposed = y_tiled - t(y_tiled)
+
+
+  v_mat = x_minus_x_transposed * char_dirs[,1] + y_minus_y_transposed * char_dirs[,2]
+  w_mat = x_minus_x_transposed * char_dirs_perp[,1] + y_minus_y_transposed * char_dirs_perp[,2]
+
+  # Use atan2 to compute the angles with respect to the characteristic
+  # directions. The angles are in the range [-pi, pi]
+  # TODO check if this atan2 has the same conventions as expected
+  angle_mat =  atan2(w_mat, v_mat)
+
+  # For each object, we then need to find the closest object in each of the
+  # four quadrants corresponding to angles (0, pi/2), (pi/2, pi), (-pi, -pi/2),
+  # (-pi/2, 0).
+  eps = 10^(-6) # used to exclude the closest object from the 1st/4th quadrant
+
+}
+
+# Implementation of the following MATLAB function
 # https://gitlab.com/fgi_nls/public/2d-registration/-/blob/main/fit_euclidean_transformation.m?ref_type=heads
 find_best_shift_hyyppa = function(pred, obs, R_local=10, k = 20, r_thresh=1.0, max_iters = 200){
   # Rename the variables for consistency with the reference code
@@ -501,6 +562,12 @@ find_best_shift_hyyppa = function(pred, obs, R_local=10, k = 20, r_thresh=1.0, m
     # TODO make this a warning
     print(msg)
   }
+
+
+  # First, we construct the feature descriptor for each object in each of
+  # the point clouds.
+  compute_feature_descriptors(xy_mat1, R_local)
+  compute_feature_descriptors(xy_mat2, R_local)
   return(c(0, 0))
 }
 
--- a/R/tree-map-alignment.R
+++ b/R/tree-map-alignment.R
@@ -646,9 +646,6 @@ find_best_shift_hyyppa = function(pred, obs, R_local = 10, k = 20, r_thresh = 1.
   xy_mat1 = pred[c("x", "y")]
   xy_mat2 = obs[c("x", "y")]
 
-  print(xy_mat1)
-  print(xy_mat2)
-
   # Centering the coordinates before matching to improve numerical stability
   xy_mat1_mean = colMeans(xy_mat1[c("x", "y")])
   xy_mat2_mean = colMeans(xy_mat2[c("x", "y")])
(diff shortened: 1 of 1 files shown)


russelldj:
I believe this is now a faithful re-implementation of the proposed algorithm. So far, I'm not super impressed with the performance. Specifically, fairly limited horizontal jitter (e.g. 1m) is enough to cause catastrophic failures. My hunch is this sensitivity is driven largely by relying only on one matching tree for the translation+orientation estimate. Especially the rotation estimation portion is very sensitive to noise, since it's only based on the location of the nearest tree.

R/tree-map-alignment.R
```diff
@@ -424,7 +429,7 @@ find_best_shift_icp = function(pred, obs) {
   pipeline_file = tempfile(fileext = ".json")
   writeLines(pipeline, pipeline_file)
 
-  cmd_call = paste0("pdal pipeline ", pipeline_file)
+  cmd_call = paste0("/home/russelldj/anaconda3/envs/pdal/bin/pdal pipeline ", pipeline_file)
```
youngdjn: This may be a remnant of some testing that should be reverted.

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+    time_taken = time_taken
+  )
 
   row.names(result) = NULL
 
   # Return the best shift
   return(result)
+}
+
+transform_points_by_homography = function(xy_mat, homography) {
```
youngdjn, line 476: Once we're better about including docstrings in this repo (or now), it would be helpful to note the expected structure of the `homography` object.

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+# Implementation of the following MATLAB function
+# https://gitlab.com/fgi_nls/public/2d-registration/-/blame/main/fit_euclidean_transformation.m?ref_type=heads#L231
+compute_feature_descriptors = function(xy_mat, R_local) {
+  xy_mat = as.matrix(xy_mat)[, 1:2]
+  # Convert the pairwise distances between each row
+  pdist = dist(xy_mat, method = "euclidean")
+  # Convert from a dist object into a matrix
+  pdist = as.matrix(pdist)
+  # Set the diagnal to the max value so it will never be the minimum
+  diag(pdist) = max(pdist)
```
youngdjn: Should it be set to the `max + 1` for the extreme case where there are only two trees to be aligned?

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  # Finding the closest neighboring object for each object
```
youngdjn: "Object" is specifically "point" for all of our possible applications, right?

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  # four quadrants corresponding to angles (0, pi/2), (pi/2, pi), (-pi, -pi/2),
+  # (-pi/2, 0).
+  eps = 10^(-6) # used to exclude the closest object from the 1st/4th quadrant
+
+  # Bounds for the quadrants
+  lower_bounds = c(eps, pi / 2, -pi, -pi / 2)
+  upper_bounds = c(pi / 2, pi, -pi / 2, -eps)
+
+  # Pre-allocating the feature descriptor matrix
+  feat_desc_mat = matrix(0, n_points, 8)
```
youngdjn: I'm often guilty of this too, but I think it's best to name the arguments, at least starting at the optional args (in this case nrow and ncol) since I don't think readers of the code will remember offhand what order they go in.

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+}
 
+
+# Implementation of the following MATLAB function
+# https://gitlab.com/fgi_nls/public/2d-registration/-/blob/main/fit_euclidean_transformation.m?ref_type=heads
+find_best_shift_hyyppa = function(pred, obs, R_local = 10, k = 20, r_thresh = 1.0, max_iters = 200) {
+  # Rename the variables for consistency with the reference code and remove unneeded columns
+  xy_mat1 = pred[c("x", "y")]
+  xy_mat2 = obs[c("x", "y")]
+
```
youngdjn: I would explicitly reference these as column names with, e.g.:

`pred[, c("x", "y")]`

as this works equally well for both matrices and data frames and is a bit more intuitive. Though how you wrote it is also perfectly legitimate.

youngdjn: Same comment applies elsewhere this occurs below

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  from_idx2_to_closest_idx1 = apply(transformed_loc_p_dist, 1, which.min)
+  from_idx2_to_min_dist1 = transformed_loc_p_dist[
+    cbind(1:nrow(transformed_loc_p_dist), from_idx2_to_closest_idx1)
+  ]
+
+  # Computing the number of matches, i.e., object pairs whose distance is
+  # below the set criterion
+  less_than_thresh = from_idx2_to_min_dist1 < r_thres
+  n_of_matches = sum(less_than_thresh)
+  return(list(from_idx2_to_closest_idx1, from_idx2_to_min_dist1, n_of_matches))
```
youngdjn: It might be slightly more intuitive to make this a *named* list so that you can reference its elements by name, and future changes to the order will not break the functions using this function's result. Syntax:

`list(name = value, name2 = value2)`

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+    # TODO make this a warning
+    print(msg)
+  }
+
+  # First, we construct the feature descriptor for each object in each of
+  # the point clouds.
+  feature_info1 = compute_feature_descriptors(xy_mat1, R_local)
+  feature_info2 = compute_feature_descriptors(xy_mat2, R_local)
+  # Extract out the components of the result
+  feat_desc_mat1 = feature_info1[[1]]
```
youngdjn: Here you could reference the list elements by name as opposed to index, if you named them when the list is created by `compute_feature_descriptors`.

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  feat_desc_pdist = cdist(feat_desc_mat2, feat_desc_mat1)
+
+  # Find the index per row with the lowest value
+  nn_indices = apply(feat_desc_pdist, 1, which.min)
+  lowest_dists = feat_desc_pdist[cbind(1:N_objects_vect[2], nn_indices)]
+
+  # Now find the second nearest index for each point in the first point cloud
+  # Mask out the first highest value.
+  # Note this must be done after extracting the distances
+  feat_desc_pdist[cbind(1:N_objects_vect[2], nn_indices)] = NaN
```
youngdjn: More commonly you'd use `NA` as opposed to `NaN`, though I think `NaN` works equivalently to `NA` in the case of numeric data types specifically.

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  # tentative matches from most reliable to least reliable
+  NNDR_vect = lowest_dists / second_lowest_dists
+  sorted_indices = sort(NNDR_vect, index.return = TRUE, decreasing = FALSE)$ix
+
+  # Then, we consider the k most promising tentative matches
+  max_n_of_matches = 0 # maximum number of matching objects
+  ## Initializing the parameters for the transformation of the best match.
+  t_best = c(0, 0)
+  theta_best = 0
+  idx_matches1_best = c() # ndices of matching objects in point cloud 1
```
youngdjn: Typo in comment

R/tree-map-alignment.R
```diff
@@ -455,16 +460,344 @@ find_best_shift_icp = function(pred, obs) {
...
+  idx_matches2_best = c() # indices of matching objects in point cloud 2
+
+  if (length(sorted_indices) < k) {
+    k = length(sorted_indices)
+    print(paste("Not enough valid correspondences, setting k to ", k))
```
youngdjn: Nice, could alternatively be a `warning`, or you could use `cat` for the combination of `print` and `paste`.

R/tree-map-alignment.R
youngdjn: Or the Tidyverse R style guide suggests not naming the *data* arguments but naming all others.

R/tree-map-alignment.R
```diff
@@ -593,10 +926,17 @@ vis1 = function(trees) {
...
 #' @export
-vis2 = function(pred, obs, obs_foc = FALSE, coords_arbitrary = FALSE, zoom_to_obs = FALSE, obs_buffer = 75) {
+vis2 = function(pred, obs, shift=c(0, 0), obs_foc = FALSE, coords_arbitrary = FALSE, zoom_to_obs = FALSE, obs_buffer = 75) {
   pred = pred |> dplyr::mutate(layer = "predicted")
   obs = obs |> dplyr::mutate(layer = "observed")
 
+  if (! identical(shift, c(0, 0))){
+    print(shift)
+    pred = pred |> dplyr::mutate(x = x + shift[1])
+    pred = pred |> dplyr::mutate(y = y + shift[2])
```
youngdjn: This could be streamlined slightly:
```
    pred = pred |> dplyr::mutate(x = x + shift[1],
                                                  y = y + shift[2])
```

youngdjn (changes requested):
Looks great! A few minor/optional comments, mostly about R syntax and conventions. I did not actually check the business logic for fidelity to the original algorithm.

R/tree-map-alignment.R
russelldj: I'm not sure I follow what that would do. This descriptor requires at least two trees in each dataset to compute the nearest tree. So if there is only one in each dataset it's expected to break.

R/tree-map-alignment.R
russelldj: Yes, I think these comments are copied from the original algorithm.

R/tree-map-alignment.R
```diff
@@ -593,10 +950,15 @@ vis1 = function(trees) {
...
 # tree maps are centered at 0,0.
 #' @export
-vis2 = function(pred, obs, obs_foc = FALSE, coords_arbitrary = FALSE, zoom_to_obs = FALSE, obs_buffer = 75) {
+vis2 = function(pred, obs, shift=c(0, 0), obs_foc = FALSE, coords_arbitrary = FALSE, zoom_to_obs = FALSE, obs_buffer = 75) {
   pred = pred |> dplyr::mutate(layer = "predicted")
   obs = obs |> dplyr::mutate(layer = "observed")
 
+  if (!identical(shift, c(0, 0))) {
+    pred = pred |> dplyr::mutate(x = x + shift[1], y = y + shift[2])
+  }
```
youngdjn: The `if` isn't necessary here, but I guess it's nice how it makes it explicit that there may actually be no shift.

youngdjn (approved):
The code looks great and seems to work as expected!

R/tree-map-alignment.R
russelldj: I'm not sure why I added that, I removed it.

PR #33, opened 2024-10-07
Add quantitative evaluation for tree registration

russelldj:
This adds functionality to compare the registration performance of different algorithms across a variety of map characteristics. Some of this functionality is more general and could be applied to other parameter tuning tasks as well.


russelldj:
My IDE made a handful of formatting changes that I accidentally committed. I can try to revert those if desired. 

I did make a few intentional changes to the tree registration code. One is updating the `find_best_shift` optimization approach to return a reformatted dataframe. Similarly, I updated the naming of the outputs in the `Hyyppa` approach. Finally, I renamed `obs_bound` -> `obs_bounds` for consistency across approaches.

vignettes/evaluate-tree-registration.qmd
```diff
@@ -0,0 +1,142 @@
...
+eval_res_threshold = function(predicted_shift, map_config, threshold = 3){
+  # Compute the distance
+  dist = eval_res_dist(predicted_shift = predicted_shift, map_config = map_config)
+  if (dist < threshold){
+    ind = 0
+  } else {
+    ind = 1
+  }
+  return(ind)
+}
```
youngdjn: Should these go into the package source code? Or is there a reason define them here on the fly?

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+  all_map_params = list()
+  # Iterate over the function arguments
+  for (name in names(simulate_tree_maps_default_args)) {
+    if (name %in% names(map_params)) {
+      # If the name is in the named list of parameters use that value
+      all_map_params[name] = map_params[name]
+    } else {
+      # Else, use the default value
+      all_map_params[name] = simulate_tree_maps_default_args[name]
+    }
```
youngdjn: This is very cool! I didn't know about the `formals` function, and I really like this approach to replacing arbitrary defaults.

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+    # If the value isn't a vector already make it a one-length vector containing that value
+    if (!is.vector(all_map_params[[name]])) {
+      all_map_params[name] = c(all_map_params[[name]])
+    }
```
youngdjn: I don't fully understand this or the need for it. Is it because you might get passed a one-length list containing a n-length vector, but you really just need the internal n-length vector? Are you concerned about this for both `name %in% names(map_params)` and `!(name %in% names(map_params))`?

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+  #' @return A dataframe where each row contains one value from each parameter drawn from the list
+
+  # Create the cross product of all elements in the list. The column names will be taken from the
+  # names in the list
+  param_grid = expand.grid(parameter_list)
+
+  # Take a random sampling of rows if requested and the number is lower than the number of rows
+  if (!is.null(n_random_samples) && n_random_samples < nrow(param_grid)) {
+    param_grid = param_grid[sample(nrow(param_grid), n_random_samples), ]
+  }
```
youngdjn: I really like this intuitive approach for either exhaustive expansion or subsampling.

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+      registration_method = registration_methods[[reg_ind]]
+      registration_method_name = registration_method_names[[reg_ind]]
+      registration_arguments = per_method_registration_arguments[[reg_ind]]
+
+      # Take the important arguments from the simulated map
+      all_registration_args = simulated_map[cbind("pred", "obs", "obs_bounds")]
+      if (!is.null(per_method_registration_arguments)) {
+        all_registration_args = c(all_registration_args, registration_arguments)
+      }
+      all_registration_args = all_registration_args[!duplicated(names(all_registration_args))]
```
youngdjn: Good call to deduplicate. I believe this would only be necessary if the args `pred`, `obs`, or `obs_bounds` were present in `per_method_registration_arguments`, which I don't think they ever should be? Might it be better to return an error if they are, rather than simply removing them?

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+
+      # EXPENSIVE LINE
+      # This is where registration actually occurs
+      predicted_shift = do.call(registration_method, all_registration_args)
```
youngdjn: Nice work finding and using this tricky R concept

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+  # Create the cross product of all parameter configurations
+  all_map_param_configurations = create_parameter_grid(all_map_params, n_random_samples = n_random_samples)
+  # Randomize the order so it's more likely that we hit errors early if they exist
+  all_map_param_configurations = all_map_param_configurations[
+    sample(nrow(all_map_param_configurations)),
+  ]
+
+  # Create names for each method if not provided
+  if (is.null(registration_method_names)) {
+    registration_method_names = as.character(seq_along(registration_methods))
```
youngdjn: Wondering if this should be left-padded with zeros using `stringr::str_pad` -- though I'm not sure a reasonable number of characters.

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+    per_method_results[[method_name]] = vector(mode = "list", length = nrow(all_map_param_configurations))
+  }
+
+  # Iterate over the parameter configurations and conduct the registration for each one
+  for (param_ind in seq_len(nrow(all_map_param_configurations))) {
+    message(paste0("iteration: ", param_ind, " / ", nrow(all_map_param_configurations)))
+    # Get one set of map parameters
+    map_params = all_map_param_configurations[param_ind, ]
+    # Create a simulated map based on these parameters
+    simulated_map = do.call(simulate_tree_maps, map_params)
```
youngdjn: This is technically simulated map**s** plural (obs and pred) but maybe keeping it singular is best, to convey it's **one set** of pred/obs as opposed to many.

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+
+  # Initialize the list of lists for the results
+  # The sub-lists for each registration method are indexed by the name of the method
+  per_method_results = list()
+  for (method_name in registration_method_names) {
+    per_method_results[[method_name]] = vector(mode = "list", length = nrow(all_map_param_configurations))
+  }
+
+  # Iterate over the parameter configurations and conduct the registration for each one
+  for (param_ind in seq_len(nrow(all_map_param_configurations))) {
```
youngdjn: Should we clarify these are **map** parameter configurations?

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+    # Get the map parameters defining the initial shift
+    if (is.null(params_df)) {
+      params = NULL
+    } else {
+      params = params_df[i, ]
+    }
+    # Compute a metric evaluating how well the predicted value matched the real one
+    metric = eval_function(results, params)
+    # Record this result
+    metric_values[[i]] = metric
```
youngdjn: Suggest making `metric_values` a vector and not a list, and then indexing into it in the final line (46) with `metric_values[i]`. I think the only reason to make it a list is if there would ever be a case where each "trial" result is represented by more than one value. But as implemented here, each trial result ("metric") is a single value.

vignettes/evaluate-tree-registration.qmd
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Define evaluation functions
+
+These are used only to visualize the results. They take the true and predicted registration values and use them to compute a scalar metric representing the quality.
+
+```{r}
+# Return the euclidian distance between the predicted and true shift
+eval_res_dist = function(predicted_shift, map_config){
+  pred_shift_vec = as.numeric(list(x=predicted_shift$shift_x, y=predicted_shift$shift_y))
```
youngdjn: I think you can simplify this as:

`pred_shift_vec = c(predicted_shift$shift_x, predicted_shift$shift_y)`,

optionally prefixing the arguments of `c` with `x=` and `y=`, but your `as.numeric` call removes those names anyway.

vignettes/evaluate-tree-registration.qmd
```diff
@@ -0,0 +1,142 @@
...
+  # TODO figure out one convention so we don't have to keep negating this
+  pred_shift_vec = -1 * pred_shift_vec
+
+  map_shift = as.numeric(list(x = map_config$shift_x, y = map_config$shift_y))
+  #cat("pred: ", pred_shift_vec, "map: ", map_shift, "\n")
```
youngdjn: Can remove this debugging code

vignettes/evaluate-tree-registration.qmd
youngdjn: It seems maybe you were setting up to accept a `predicted_shift` parameter with multiple rows of predicted shifts, but I don't think this code exactly accommodates that, nor does that situation arise in the current implementation.

R/parameter_tuning.R
```diff
@@ -0,0 +1,218 @@
...
+  #' @param smoothing_bandwidth Width of the kernel used to smooth the trend line. May be a scalar or one value per attribute being visualized.
+  #' @param title Title of the plot
+
+
+  # Compute the metric values for each row in the results
+  metric_values = compute_metric_per_trial(
+    results_df = results_df,
+    eval_function = eval_function,
+    params_df = parameter_df
+  )
```
youngdjn: Because the evaluation is being done per trial, the `eval_res_threshold` function always returns either 0 or 1, so it does not yield any insight beyond what `eval_res_dist` returns. For that to be useful, I think this evaluation would have to process multiple "trials" (random map generations using the same parameters) at once so that a cross-trial "success rate" could be computed.

youngdjn (changes requested):
This is awesome! No worries on the reformat; that's the style I eventually want to apply to the whole package. Mostly some nitpicky inline comments, and a few overarching ones:

- "trials" had a different meaning than I had in my head. I believe you're using "trials" to refer to map configurations. I had thought of it as random instances of a given map configuration. I realize it's a generic term that could be used for either, but maybe it's worth being more specific and calling it "map configurations" in your case?
- Relatedly, after this is merged, it would be nice to accommodate testing multiple random instances of a given map configuration, so that a success rate other than 0 or 1 can be determined. Not critical at this stage, but keeping this in mind now may help accommodate this later.
- Thanks for creating docstrings. I thought they were supposed to go immediately above the function ([per this source](https://r-pkgs.org/man.html#sec-man-workflow)) but your in-function docstrings seem to be working fine!
- I think we should standardize around calling this task "alignment" or "registration". Prior to this code, I'd been calling it "alignment". Do you think we should avoid that?
- Please name the source code file with dashes instead of underscores, unless a higher-level separator is also needed.

Future enhancements (after merging):

- Accommodate multiparameter/interaction visualization
- We should think about parallelizing and at what level that would be most efficient. The grid search is already (optionally) parallel. But its current alternative (optimizer) is not parallelizable.

vignettes/evaluate-tree-registration.qmd
russelldj: Good point. I was initially thinking the user might want to define their own custom metric. But it's probably better to provide at least a few basic ones in the package source code.

R/parameter_tuning.R
russelldj: Yeah, it seems effective. The one issue I had with it involved a negative default being interpreted in a strange way I couldn't understand. I think it may have been split into a negative sign and the value or something.

R/parameter_tuning.R
russelldj: This is for when you pass a scalar value. Since we're doing a cross product of all the potential values for each parameter with `expand.grid`, I assumed each set of potential values had to be represented as a list, even if it only had one value. The `expand.grid` function might directly support some attributes being scalars, but the documentation is not clear.

R/parameter_tuning.R
russelldj: Smart, I agree the multi-trial approach would be better. I'd argue though that you get something close to a success rate using the kernel-smoothed trend line, since it's averaging the `0/1` values. But the individual data points are not particularly informative.

youngdjn: OK fair point -- fine to keep as is for now!

youngdjn (approved):
This looks fantastic now--thank you!

PR #34, opened 2024-11-07
Support new platforms by reorganizing how metadata is standardized

russelldj:


R/metadata-extraction_imagery_dataset.R
```diff
@@ -21,18 +20,17 @@ extract_flight_speed = function(exif) {
...
-
 }
 
 # Get a polygon of the mission
 # image_merge_distance: The horizontal distance between images below which they are merged into one
 # mission polygon. Keep only contiguous patches > min_contig_area.
 #' @export
 extract_mission_polygon = function(exif, image_merge_distance, min_contig_area = 1600) {
-
-  exif = sf::st_transform(exif, 3310)
```
russelldj, line 33: Add a conversion to UTM, but make that function check if it's a no-op.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,15 +26,85 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
+    "Matrice 300"
+  )) {
+    return(extract_imagery_perimage_metadata_DJI(exif))
+  } else {
+    print(paste0("Aircraft model name \'", aircraft_model_name, "\' not supported"))
+    return(NULL)
+  }
+}
+
+print(length(exif_list))
```
russelldj: Since prep-exif isn't used anymore, we still need to visualize, sort by time, and remove rows without GPS at this point after the format has been standardized.

russelldj:
Documentation cleanup is still required. Some fields don't seem to be parsing on the per-image level such as the white balance. Others such as the terrain following seem broken at the dataset level.

R/metadata-extraction_imagery_perimage.R
```diff
@@ -118,11 +118,15 @@ extract_rtk_fix = function(exif) {
...
+  exif = sf::st_drop_geometry(exif)
+  # This field may not be actually present in the data but only introduced by
+  # row_binding across multiple platforms
+  if ("GPSXYAccuracy" %in% colnames(exif) && !all(is.na(exif$GPSXYAccuracy))) {
     accuracy_x = exif$GPSXYAccuracy
     accuracy_y = exif$GPSXYAccuracy
-  } else {
+  } else{
+    # Try extracting the deviation from the RTK field
+    # TODO should we instead prioritize the RTK value if both are present?
```
russelldj, line 129: Is there a situation where both the `GPSXYAccuracy` and `RtkStdLon/RtkStdLat` fields would be set?

R/metadata-extraction_imagery_dataset.R
```diff
@@ -21,19 +21,25 @@ extract_flight_speed = function(exif) {
...
 # image_merge_distance: The horizontal distance between images below which they are merged into one
 # mission polygon. Keep only contiguous patches > min_contig_area.
 #' @export
-extract_mission_polygon = function(exif, image_merge_distance, min_contig_area = 1600) {
-
-  exif = sf::st_transform(exif, 3310)
-  ptbuff = sf::st_buffer(exif, image_merge_distance)
-  polybuff = sf::st_union(ptbuff)
-  poly = sf::st_buffer(polybuff, -image_merge_distance + 1) |> sf::st_cast("MULTIPOLYGON")
+extract_mission_polygon = function(exif, image_merge_distance, min_contig_area = 1600, boundary_method = "concaveman") {
```
russelldj: Note that #41 should be merged before this PR. This branch was rebased onto that one so it contains both sets of changes.

R/metadata-extraction_imagery_dataset.R
```diff
@@ -154,31 +161,34 @@ extract_flight_terrain_correlation = function(exif) {
...
-  ground_elev_core = ground_elev[agl > agl_lwr & agl < agl_upr]
+  agl_core_mask = agl > agl_lwr & agl < agl_upr
+
+  drone_altitude_core = drone_altitude[agl_core_mask]
+  ground_elev_core = ground_elev[agl_core_mask]
 
   # Get the correlation between the altitude of the drone and the ground elevation (i.e. trerrain
   # follow tightness)
-  flight_terrain_correlation = cor(exif_elev_core, ground_elev_core) |> round(2)
+  # Note that for some small regions the DEM will not have any variation so this
```
russelldj: Is there a better way to handle this?

russelldj:
I think this PR is complete and ready for review.

R/metadata-extraction_imagery_dataset.R
youngdjn: Maybe in these edge cases, force cor to 1 using something like this (untested):

```R
ground_elev_no_variation = length(unique(ground_elev_core)) == 1
if (ground_elev_no_variation) {
  flight_terrain_correlation = 1
} else {
  flight_terrain_correlation = cor(exif_elev_core, ground_elev_core) |> round(2)
}
```

I'm OK forcing it to 1 because the main purpose is to be able to filter out missions with poor terrain following but that would be irrelevant in these no-variation cases.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,59 +28,74 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
-                        input_type = "dataframe",
-                        plot_flightpath = FALSE,
-                        crop_to_contiguous = TRUE,
-                        min_contig_area = 1600,
-                        .options = furrr_options(seed = TRUE))
-
-metadata_list = map(res, ~.x$dataset_metadata)
-polygon_list = map(res, ~.x$mission_polygon)
-images_retained_list = map(res, ~.x$images_retained)
+summary_statistics = furrr::future_map(metadata_per_sub_dataset, extract_imagery_dataset_metadata)
```
youngdjn: Could you add a comment above this like:
`# Use the standardized image-level metadata extracted in previous step to compute dataset-level summary statistics`

R/metadata-extraction_imagery_dataset.R
```diff
@@ -1,17 +1,17 @@
+library(concaveman)
 # Functions to extract metadata from a drone imagery dataset, using an exif dataframe that has
 # already been prepped using prep_exif.
```
youngdjn: Maybe this comment could be expanded with a little more guidance on the expectations and intended use. Also this is no longer expecting an exif data frame, but instead an extracted & standardized per-image metadata dataframe, correct?

Maybe the comment could include (or consist entirely of) something like:
`# Functions to take image-level metadata that has already been assigned standardized column names (regardless of drone platform) and compute summary metrics at the dataset level.`

R/metadata-extraction_imagery_dataset.R
youngdjn, line 1: Should the `exif` argument of all of the attribute-specific functions be renamed to something like `metadata`? It is no longer direct-from-image EXIF data.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,59 +28,74 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
+  unique_sub_missions,
+  function(sub_mission_ID) {
+    # Extract the exif rows matching that sub-mission ID
+    sub_mission_exif <- exif |>
+      filter(submission_id == sub_mission_ID)
+    return(sub_mission_exif)
+  }
+)
+
+# Compute the corresponding baserow data for each of the sub-missions
```
youngdjn: This is a very general comment for doing something very specific: getting the aircraft model name. Could the comment be made more specific, or are you anticipating future expansion to include other attributes? I'm thinking of something like:

`# Get aircraft model name to use for determining the appropriate image-level data extractor functions to use`

R/metadata-extraction_imagery_perimage.R
```diff
@@ -321,16 +347,26 @@ extract_altitude_asl = function(exif, candidate_asl_columns = c("AbsoluteAltitud
...
+    "Matrice 210 RTK",
+    "Matrice 100",
+    "Matrice 300",
+    "eBee X" # As far as I can tell, there is no issue with this using the DJI parser
+  ))) {
+    stop(paste("Platform ", platform_name, " is not supported"))
   }
 
+  # In the future, some of these parsing steps might be dependent on the platform but for now it's
+  # not required
```
youngdjn: Could you expand this comment to explain how you envision this might happen in the future? Would it be like:

`This future approach could involve classifying the platform_name into relevant exif format groups (e.g. group_a and group_b) and then calling the relevant group-specific attribute extraction functions below. E.g. extract_lon_lat_group_a vs extract_lon_lat_group_b. Note that for some attributes, it may not be necessary to specify alternatives by group even if it is for others.`

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,59 +28,74 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
+    aircraft_model_name = baserow_for_sub_mission$aircraft_model_name
+
+    return(aircraft_model_name)
+  }
+)
+
+# Extract the metadata in a standardized manner no matter the platform
+metadata_per_sub_dataset = furrr::future_map(seq_along(exif_per_sub_mission), function(i) {
+  extract_imagery_perimage_metadata(exif = exif_per_sub_mission[[i]], platform_name = aircraft_model_names[[i]])
 })
```
youngdjn: For tractability, you could use `future_map2` (or the generalized case for n parallel arguments `future_pmap`) here instead of defining the anonymous function. Not essential to change.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,59 +28,74 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
+
+    # Extract the aircraft model name
+    aircraft_model_name = baserow_for_sub_mission$aircraft_model_name
+
+    return(aircraft_model_name)
+  }
+)
+
+# Extract the metadata in a standardized manner no matter the platform
+metadata_per_sub_dataset = furrr::future_map(seq_along(exif_per_sub_mission), function(i) {
```
youngdjn: For `furrr` to actually run this in parallel, you have to call the following somewhere before this in the script:
`future::plan(multisession)`. Without this it falls back to serial and works identically as the corresponding functions in the `purrr` package. Parallelizing here does not seem to be critical because it is quite fast in serial.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
youngdjn: This call, however, could be substantially sped up by enabling parallelization, mostly because it involves downloading a DEM for every sub-mission.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
```diff
@@ -25,59 +28,74 @@ crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PRO
...
-polygon_list = map(res, ~.x$mission_polygon)
-images_retained_list = map(res, ~.x$images_retained)
+summary_statistics = furrr::future_map(metadata_per_sub_dataset, extract_imagery_dataset_metadata)
+print("Finished processing dataset-level summary statistics")
 
-metadata_perdataset = bind_rows(metadata_list)
-polygon_perdataset = bind_rows(polygon_list)
-images_retained = unlist(images_retained_list)
+# Extract the elements of the summary statistics
+metadata_perdataset = bind_rows(map(summary_statistics, ~ .x$dataset_metadata))
```
youngdjn: Looks like you can replace `~ .x$dataset_metadata` with `"dataset_metadata"`. Applies to the lines below this too. I think I was the one who wrote this in the original syntax though, sorry.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
youngdjn: The name `metadata_per_sub_dataset` to me implies that it is sub-dataset-level summary. Could this be renamed to indicate that it is image-level? Also, I don't think we really have a "sub-dataset" concept. I was trying to use "dataset" as a generic term for both "mission" and "sub-mission". So maybe rename to something like `image_level_metadata_per_dataset`.

youngdjn: The printed message below could be updated to match this terminology too.

R/metadata-extraction_imagery_perimage.R
```diff
@@ -118,11 +118,15 @@ extract_rtk_fix = function(exif) {
 #' @export
 
 extract_accuracy = function(exif) {
-  exif = st_drop_geometry(exif)
-  if ("GPSXYAccuracy" %in% colnames(exif)) {
+  exif = sf::st_drop_geometry(exif)
+  # This field may not be actually present in the data but only introduced by
```
youngdjn, line 122: Good catch!

R/metadata-extraction_imagery_perimage.R
youngdjn: I tried but I could not find any EXIF that had `GPSXYAccuracy` at all! I'm not sure where this came from. For non-RTK, the best potential option I see is `circleofconfusion` and it can exist along with RtkStdLon and be different. I'm not sure what to do about this but I think for now it's OK to focus on reporting the RTK accuracy if RTK and no accuracy otherwise. Accuracy for instantaneous point coordinates is pretty tenuous anyway I think.

R/metadata-extraction_imagery_dataset.R
```diff
@@ -45,31 +51,31 @@ extract_mission_polygon = function(exif, image_merge_distance, min_contig_area =
...
       warning(n_polys, " non-contiguous image clusters in dataset ", exif$dataset_id[1], ". Retaining all becaus all have area > ", min_contig_area, " m^2.")
     }
-
   } else if (n_polys == 0) {
     stop("No contiguous images in dataset ", exif$dataset_id[1])
   }
 
 
-  polysimp = sf::st_simplify(poly, dTolerance = 10) |> sf::st_cast("MULTIPOLYGON")
+  polysimp = sf::st_simplify(poly, dTolerance = 1) |> sf::st_cast("MULTIPOLYGON")
```
youngdjn: What is the reasoning behind this change? Is this the best value to use also if the method is specified as `dilate_erode`?

R/metadata-extraction_imagery_dataset.R
```diff
@@ -80,7 +86,7 @@ extract_camera_pitch_summary <- function(exif) {
   # just in case the pitch is reported as negative for backward (even thouth it apparently never
   # is), so that for smart-oblique missions, the low quantile represents nadir and the high quantile
   # represents the oblique mission pitch
```
youngdjn: Could you move this comment about adjusting for nadir to the image-level extraction where the +90 happens? I realize I didn't comment it there, but no this comment here is irrelevant.

youngdjn (changes requested):
This looks good! Just a few minor requests, mostly for updated comments. Once the changes are made (assuming you agree with them) you can go ahead and merge.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
youngdjn: However in testing I noticed that (only when run with parallelization there is an error like `Cannot find function st_zm` which is really confusing me because every call to st_zm in the `ofo-r` codebase is prefaced correctly with `sf::`. So it must be a third-party function calling st_zm but it will be really hard to debug because it only happens with parallelization enabled.

R/metadata-extraction_imagery_perimage.R
russelldj: It looks like the eBee reports `GPSXYAccuracy` but I didn't see it for the DJI platforms. However, all of them report the RTK metric for the NRS data I was looking at.

I also saw `circleofconfusion` but I don't think it's what we want. It appears to be an optics term related to the magnitude of the imaging blur: [wiki](https://en.wikipedia.org/wiki/Circle_of_confusion). This is supported by the fact that the values are sub-millimeter.

R/metadata-extraction_imagery_dataset.R
russelldj: Sorry, this was a one-off and I shouldn't have committed it. I got an issue with some small flight polygons being entirely eliminated by the value of 10 so I decreased it. Maybe there's a better way to preserve small polygons. 

Is this simplification primarily for computational reasons or to make the result look good? If it's computational, I wonder if a smaller simplification threshold such as 1 would still result in fast enough operations using the resultant polygon.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
russelldj: Hmm, that is weird. Do you think multiple threads can download the DTM at the same rate as only one thread, or would they be fighting for network bandwidth?

russelldj:
I was doing my final tests to ensure everything is working and the process of downloading the DEM has gotten dramatically slower, even when reverting to the sequential version. I'm trying to debug this now.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
russelldj: There was a single instance of `st_zm` that wasn't prefixed in `R/tree-detection-accuracy-assessment.R`. I fixed that and the error went away.

russelldj:
> I was doing my final tests to ensure everything is working and the process of downloading the DEM has gotten dramatically slower, even when reverting to the sequential version. I'm trying to debug this now.

This was a simple indexing error farther up in the script. It was trying to run all of the images as one dataset, so the AOI that was downloaded was huge.

russelldj:
I tried to remove the concaveman commit but it didn't work out, so I just made it the non-default option. Otherwise, this all seems to be working, with the exception of eliminating some of the small polygons because of simplification.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-sub-mission.R
youngdjn: Ah good find on the one `st_zm`. I think multiple threads (up to some number like 8?) will each have the download speed of a single serial thread, so parallelizing should speed it up substantially. Also there appears to be a lot of per-DEM overhead (not sure what -- waiting for the API to respond? -- that is not bandwidth-related and should scale pretty linearly with number of threads.

R/metadata-extraction_imagery_dataset.R
youngdjn: It was more because I thought it was the best at capturing the polygon we actually designed the mission to represent, rather than (at finer scale) the idiosyncrasies of how it was actually collected.

However that may not be good logic (maybe we do want to capture the actual, not intended, polygon), and I didn't know that 10 could entirely eliminate the polygon, so I'd support using a smaller value.

youngdjn:
> This was a simple indexing error farther up in the script. It was trying to run all of the images as one dataset, so the AOI that was downloaded was huge.

Good catch, that was tricky! That is a problem with the way `dplyr` simplifies dataframe operations: if a given name exists as a variable in the environment *and* as a column in the subject data frame, it always gives priority to the data frame column, on both the left and right side of operators like `==`.

Thanks for making all these changes and merging! It's great to have this incorporated.

I guess we still need to do something about the coarse polygon simplification eliminating the polygon. I just added this issue #42 .

russelldj:
Thanks for the explanation, I actually assumed the other option, that the local variable took precedent over the column name, so that's helpful to know.

PR #36, opened 2024-11-19
WIP: Convert to cloud optimized format and create thumbnails

russelldj:
This has code to take a folder of photogrammetry outputs and turn them into a more standardized format. Specifically, raster and point cloud data is cropped to the flight polygon bounds and exported in a cloud optimized format. Then, a thumbnail is generated for the raster data as a png. 

I opened this to make sure we were up to date on the changes to this branch. A minor detail is that columns need to be renamed after they are compiled in `list_photogrammetry_outputs`. Also, we should consider subsetting the meshes to the desired regions. Finally, the code should be cleaned up, including format, better documentation, and consistent logging/print outs.

### DIFF
--- a/.Renviron
+++ b/.Renviron
@@ -1,0 +1,1 @@
+TMP = '/ofo-share/tmp'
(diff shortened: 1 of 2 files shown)


amrithasp02 (approved):

youngdjn:
One thing to add still is creation of CHMs, ideally using `ofo::chm_from_coregistered_dsm_dtm`, probably with `res = 0.25`.

Also, upstream of everything we may want to first run `ofo::transform_to_local_utm` in case the photogrammetry outputs were in lon/lat or some other non-standard CRS. This carries the consequence of data loss (smoothing) due to reprojection, so it may not be ideal and I'm interested in thoughts.

PR #39, opened 2024-11-27
Fix DTM-DSM logic errors in CHM computation

russelldj:
There were several mistakes in the logic around creating the CHM caused by mixing up the DSM and DTM variables. The final data was produced with these fixes.


youngdjn (approved):
Good catch! This fix looks good. Thanks.

PR #41, opened 2024-12-06
Add option to use concaveman for extracting mission polygons

russelldj:
This adds the concave hull method implemented in `concaveman` as an alternative default method to compute the flight polygons. Currently it uses buffering and erosion. This seems to high-quality results that match the intuition for what a reasonable boundary is. I changed the concavity parameter from the default of 2 to 4 to encourage slightly more simplified shapes. 

plot

I still need to test this function before it's ready to merge.

### DIFF
--- a/R/metadata-extraction_imagery_dataset.R
+++ b/R/metadata-extraction_imagery_dataset.R
@@ -1,3 +1,4 @@
+library(concaveman)
 # Functions to extract metadata from a drone imagery dataset, using an exif dataframe that has
 # already been prepped using prep_exif.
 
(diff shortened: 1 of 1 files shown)


R/metadata-extraction_imagery_dataset.R
```diff
@@ -1,3 +1,4 @@
+library(concaveman)
```
youngdjn, line 1: Rather than adding dependencies to a function definitions file, they should be added to the `Imports` list in the `DESCRIPTION` file.

R/metadata-extraction_imagery_dataset.R
```diff
@@ -21,19 +22,25 @@ extract_flight_speed = function(exif) {
...
-  poly = sf::st_buffer(polybuff, -image_merge_distance + 1) |> sf::st_cast("MULTIPOLYGON")
+extract_mission_polygon = function(exif, image_merge_distance, min_contig_area = 1600, boundary_method = "concaveman") {
+  # TODO we could check that it's a meters-based projected CRS
+  if (boundary_method == "concaveman") {
+    poly = concaveman::concaveman(exif, concavity = 4)
+    poly = poly |> sf::st_cast("MULTIPOLYGON")
+  } else if (boundary_method == "dilate_erode") {
+    ptbuff = sf::st_buffer(exif, image_merge_distance)
+    polybuff = sf::st_union(ptbuff)
+    # This line is dangerous for single lines of images since
```
youngdjn, line 39: Could you finish this comment?

youngdjn (changes requested):
Concaveman as implemented here appears to be working great for missions with contiguous points. I agree the `concavity` value of `4` appears to work well. Unfortunately though, concaveman does not appear to deal well with cases of non-contiguous points. We need to tolerate this possibility because during initial image curation, it is possible we will encounter (non-ideal) datasets that have multiple non-contiguous groups of points. Here is how it dealt with one situation:

image

You can reproduce this with the following script:

```R
devtools::load_all()
library(tidyverse)
library(sf)

exif = read_csv("/ofo-share/drone-imagery-organization/3b_exif-unprocessed/exif_2019-focal.csv")

ids = unique(exif$mission_id)

exif_foc = exif |>
  filter(mission_id == ids[1]) |>
  prep_exif()

exif_foc = exif_foc[c(1:100, 1001:1100), ]

poly = extract_mission_polygon(exif_foc)

plot(exif_foc |> st_as_sfc())
plot(poly, add = TRUE, lwd = 4)
```
Do you have any ideas on how to deal with this cleanly?

The concaveman method can also end up with a lot of legitimate border images falling outside of the generated polygon -- far more than the dilate-erode method. These would get dropped by the `crop_to_contiguous` step (which is enabled by default) in `extract_imagery_dataset_metadata` which we don't want. One way around this is to figure out how far beyond the generated boundary a legitimate image might fall, and adjust the buffer in [this line](https://github.com/open-forest-observatory/ofo-r/blob/0ccd0a45ded201d68b88e4d1fa34b8a08e9470fe/R/metadata-extraction_imagery_dataset.R#L517) to use that distance.

R/metadata-extraction_imagery_dataset.R
russelldj: Thanks, I was confused why there weren't any other imports.

russelldj:
Thanks for doing these tests, I only checked a couple of the real datasets so I completely missed these issues. I can mess around with some of the parameters to see if I can get better results, but I am surprised it did so poorly with two widely-separated clusters of points. Maybe the concavity needs to be set lower.

I'm not quick sure why points are falling outside of the polygon. My guess would be this is because of the simplification step called later. If we set the buffer in the image-subsetting step to the same value as the simplification tolerance, it should ensure that any points that were in the original (un-simplified) polygon are still retained.

russelldj:
closing because this functionality was added in #57.

PR #44, opened 2024-12-17
Exclude datasets that are empty after filtering

youngdjn:
If the dataset is so small that the simplified geospatial polygon representing the dataset is empty, then no images will be selected which will cause an indexing error. This PR is one potential solution to this problem. What do you think @russelldj ?


russelldj (approved):
Looks good to me. We just may have to do some additional checks in later steps since we can no longer assume that the summary statistics metadata will be present for a given dataset.

PR #45, opened 2024-12-17
Update base metashape config for new version of automate-metashape

russelldj:
The config needed to be updated to be consistent with the most recent version of `automate-metashape`
- Add an `addPhotos` collection
- Add an `exportCameras` collection
- Add `secondary_photo_path`
- Duplicate the ortho surface so it is not coerced into a single-length list. See issue [here](https://github.com/vubiostat/r-yaml/issues/69).

I have tested this config piecemeal, I'm currently running a job start-to-finish to validate it.

### DIFF
--- a/sandbox/drone-imagery-ingestion/full-run-configs/base.yml
+++ b/sandbox/drone-imagery-ingestion/full-run-configs/base.yml
@@ -10,6 +10,7 @@ load_project: ""
 # If there are multiple photo folders, set path to the folder that contains all the photo folders
 # If there are no photos to add (e.g., this is an existing project that already has photos in it, set to an empty string ("")
 photo_path: "/example/path/to/photo/folder"
+photo_path_secondary: ""
 separate_calibration_per_path: True
 multispectral: False # Is this a multispectral photo set? If RGB, set to False.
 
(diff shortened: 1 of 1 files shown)


russelldj:
The full metashape run using these derived configs completed successfully.

sandbox/drone-imagery-ingestion/full-run-configs/base.yml
```diff
@@ -129,7 +142,16 @@ buildDem: # (Metashape: buildDem, (optionally) classifyGroundPoints, exportRaste
...
+    # TODO This is a very annoying hack because the R yaml parser simplifies one-length lists into a
+    # single element. To get around this, we duplicate the element, running the same operation
+    # twice. Fortunately, this operation is relatively quick in the scheme of things.
+    # See discussions below.
+    # https://stackoverflow.com/questions/23475743/yaml-in-r-passing-in-a-one-element-list
+    # https://github.com/vubiostat/r-yaml/issues/69
+    # Options: "DTM-ptcloud", "DSM-ptcloud", "DSM-mesh", and/or "Mesh". The surface to build the orthomosaic onto. DTM and DSM refer to elevation models built by Metashape and must be configured to be computed via buildDem, above. Mesh refers to using the mesh model directly rather than first computing a DEM.
+    surface:
+        - DSM-mesh
+        - DSM-mesh
```
youngdjn: I see this is duplicated as part of a necessary hack. However maybe we could also generate a `DSM-ptcloud` (I think these actually look better, something I realized after running the other projects with only `DSM-mesh`)

youngdjn: As a side comment, is it a problem that this list gets simplified to a single element? Does that break automate-metashape?

youngdjn (approved):
Looks good, just one minor suggestion for the processing (not the implementation) and question.

sandbox/drone-imagery-ingestion/full-run-configs/base.yml
russelldj: That sounds like a good alternative, I was just trying to keep the outputs consistent with the past. Yes, the issue is with automate-metashape. If surface(s) are represented as a single string, it tries to iterate over the characters in that string and looks for the surface `D`.

PR #52, opened 2025-01-09
Only correct the pitch for DJI platforms

russelldj:
We were applying a +90 degree pitch correction to data from all platforms in an attempt to match our "degrees up from nadir" convention. However, the Autel and eBee platforms already report metadata consistently with our format so they do not need to be corrected. Since we have access to the platform when standardizing the metadata it's easy to apply the correction only to the DJI platforms.


youngdjn (approved):
This looks great! Thank you.

PR #55, opened 2025-01-13
Remove prep_exif from namespace

russelldj:
This function had been removed from the package but is still included in the namespace file


youngdjn:
Good catch. Did you remove this by running `devtools::document()` or by editing the file directly? It says not to edit by hand but I'm not sure exactly what problems that might introduce, or how it is different from `devtools::document()`.

russelldj:
I did this by hand. I just ran `devtools::document()` on main and it removed that line while also adding two more exports, `extract_image_frequency` and `extract_original_file_name`. It also updated a bunch of `man` files. Once the github issues (https://www.githubstatus.com) are resolved I can run `devtools::install()` on this branch and commit the results.

youngdjn:
OK thank you! I assume by `devtools::install()` you mean `devtools::document()` and that sounds good! Feel free to push that directly to main.

russelldj:
> OK thank you! I assume by `devtools::install()` you mean `devtools::document()` and that sounds good! Feel free to push that directly to main.

Will do. Yes, sorry, that was a typo.

russelldj:
Closed with fe227f1

PR #57, opened 2025-01-14
Reorganize the ingest

russelldj:
This contains a few changes trying to streamline and debug the dataset ingest.


russelldj:
This adds a variety of changes to the drone imagery ingest pipeline.
- Remove `dataset_id_image_level` column from the image-level data. This is now replaced by two more explicit columns, `mission_id` and `sub_mission_id` for image-level data. 
- The concept of per-image metadata specific to either a mission or sub-mission has been removed. Now image-level metadata is unified across the two.
- `dataset_id` is retained for summary content at the mission or sub-mission level, representing the corresponding `mission_id` or `sub_mission_id`.
- `submission_id` has been replaced by `sub_mission_id` in imagery-derived metadata. This is because `submission_id` is used in baserow to denote an imagery submission, a sub-unit within a project.
- Since the metadata parsing is dependent on the platform, standardizing the baserow metadata is moved earlier in the process.
- To avoid redundancy, the per-image metadata is computed only once and saved. Then the summary statistics at the mission or sub-mission level are computed as a later step. 
- Minor documentation and robustness is updated throughout.

R/web-catalog-creation_drone-imagery-curation.R
```diff
@@ -559,28 +560,28 @@ make_mission_details_pages = function(
     mission_details_page_dir,
     published_data_path = "",
     data_server_base_url = "") {
-
-  mission_summary = mission_summary |>
-    dplyr::arrange(dataset_id)
+  # Sort the summary statistics by dataset_id (mission_id)
+  mission_summary = mission_summary |> dplyr::arrange(dataset_id)
```
youngdjn: Do you think it would be appropriate or useful to add a comment like:

> This assumes the column `dataset_id` has been added to the `mission_summary` dataframe by the function that creates it (ultimately, *OK I actually could not easily track this back all the way to where it's added, is it `compute_summary_statistics` defined in script 08?*), representing either the mission or sub-mission, whichever is currently being processed.

But also I realized that this is only being run (at least currently) for missions, not sub-missions. Do you think we should keep the `dataset_id` concept here (as you are) because it could theoretically in the future be run for sub-missions too? But in that case, should we rename this function `make_dataset_details_pages`? I don't really know, but interested in your opinion since you've thought more about this recently.

Also how bad is it that the function here is in the ofo-r source code, but the function that preps the data for it is defined in a workflow script in sandbox? should `compute_summary_statistics` be moved into the oro-r source code and maybe renamed? Very open to opinions. I realize you made this function out of what I had previously written as loose script which is a big improvement so I don't mean for this to fall on you.

R/web-catalog-creation_drone-imagery-curation.R
```diff
@@ -559,28 +560,28 @@ make_mission_details_pages = function(
...
 
-  for (i in 1:ndatasets) {
+  # Rename mission_id to dataset_id
+  mission_points = mission_points |> dplyr::rename(dataset_id = mission_id)
 
-    dataset_id_foc = dataset_ids[i]
+  for (i in 1:ndatasets) {
+    # Get a single row from the summary statistics
+    dataset_id_foc = dataset_ids[[i]]
+    # Extract the image-level metadata that's associated with that dataset
```
youngdjn: Comment should say mission-level, not image-level I think.

sandbox/drone-imagery-ingestion/05_prep-metashape-configs.R
```diff
@@ -43,12 +43,10 @@ for (dataset in datasets) {
...
 
-scenarios
-
 make_derived_configs(
   base_yaml_filepath = BASE_YAML_FILEPATH,
   scenarios,
   derived_yaml_out_folder = derived_yaml_out_folder,
   automate_metashape_path = AUTOMATE_METASHAPE_PATH,
-  n_shell_splits = N_SHELL_SPLITS
+  n_shell_splits = min(N_SHELL_SPLITS, nrow(scenarios)) # Make sure we don't request more splits than config files
```
youngdjn, line 51: Good catch!

sandbox/drone-imagery-ingestion/06_read-folder-exif-to-csv.R
```diff
@@ -38,16 +38,16 @@ future::plan("multisession")
...
+# Add the mission ID and sub_mission ID to the EXIF data, assuming that the folder organization is:
 # {any abs path}/<mission_id>/<submission_id>/<incrementing number per 10000 images>/<image>.jpg
 
-submission_path = dirname(dirname(image_paths))
-submission_id = basename(submission_path)
-mission_path = dirname(submission_path)
+sub_mission_path = dirname(dirname(image_paths))
+mission_path = dirname(sub_mission_path)
+
+sub_mission_id = basename(sub_mission_path)
```
youngdjn, line 47: Thanks for improving this consistency

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
```diff
@@ -0,0 +1,103 @@
...
+
+# For parallelizing, make a list of subsets of the exif dataframe, one for each sub-mission
+exif_per_sub_mission <- lapply(
+  unique_sub_mission_ids,
+  function(unique_sub_mission_id) {
+    # Extract the exif rows matching that sub-mission ID
+    sub_mission_exif <- exif |>
+      filter(sub_mission_id == unique_sub_mission_id)
+    return(sub_mission_exif)
+  }
```
youngdjn: I generally prefer to avoid anonymous functions but this one is short enough that it's OK

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
```diff
@@ -0,0 +1,103 @@
...
+    # There should only be one matching row
+    if (nrow(baserow_for_sub_mission) != 1) {
+      stop(paste("Error: there was not one corresponding baserow entry: ", baserow_for_sub_mission))
+    }
+
+    # Extract the aircraft model name
+    aircraft_model_name = baserow_for_sub_mission$aircraft_model_name
+
+    return(aircraft_model_name)
+  }
```
youngdjn: This one's pretty long. Do you have a strong preference for this or what do you think about defining the function just above, or at the top of the script?

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
youngdjn, line 1: Annoyingly (not your fault) I'm unable to see the diffs in this file due to the rename. But overall it looks good.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
```diff
@@ -0,0 +1,103 @@
...
+  .f = extract_imagery_perimage_metadata,
+  .progress = TRUE,
+  .options = furrr::furrr_options(seed = TRUE)
+)
+# Bind the rows together across all sub-missions
+metadata_perimage = bind_rows(metadata_per_sub_mission)
+
+# Copy the sub-mission and mission ID to the output metadata
+metadata_perimage$sub_mission_id = exif$sub_mission_id
+metadata_perimage$mission_id = exif$mission_id
```
youngdjn, line 111: What if there was a missing baserow entry for one sub-mission, or the entry existed but had a bad aircraft platform type so we couldn't process the EXIF, and we wanted to tolerate (skip) that rather than error? In that case there would not be a 1:1 match between `metadata_perimage` and `exif`, right? Is this likely or important enough that we should address it here?

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn, line 1: I think the top comment needs to be updated. Also, should the file be renamed to include sub_mission since now it does? (Though I'm glad you didn't rename it yet because it allows me to see the diffs.)

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -27,93 +27,98 @@ EXIF_PATH = "/ofo-share/drone-imagery-organization/3b_exif-unprocessed/"
...
+
+# Output files per mission or sub-mission
+metadata_per_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-exif-metadata_perdataset_", IMAGERY_PROJECT_NAME, ".csv"))
+mission_polygons_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-polygons_", IMAGERY_PROJECT_NAME, ".gpkg"))
+
+metadata_per_sub_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("sub-mission-exif-metadata_perdataset_", IMAGERY_PROJECT_NAME, ".csv"))
+sub_mission_polygons_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("sub-mission-polygons_", IMAGERY_PROJECT_NAME, ".gpkg"))
+
+# Functions
+compute_summary_statistics = function(
```
youngdjn: As mentioned in a previous comment, I'm wondering if this should be moved into the source code.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -27,93 +27,98 @@ EXIF_PATH = "/ofo-share/drone-imagery-organization/3b_exif-unprocessed/"
...
-purrr::walk(folders,
-            create_dir)
-
-write_csv(metadata_perimage, metadata_perimage_filepath)
-write_csv(metadata_perdataset, metadata_perdataset_filepath)
-st_write(polygon_perdataset, polygons_filepath, delete_dsn = TRUE)
+# Read in image-level metadata that was parsed in step 07
+image_metadata = read_csv(metadata_perimage_input_filepath)
+
+# Compute the summary statistics based on missions
```
youngdjn: Since the return value is not the main purpose of the function (but instead saving files is), can you please include "and save to the provided file paths" to the comment?

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -27,93 +27,98 @@ EXIF_PATH = "/ofo-share/drone-imagery-organization/3b_exif-unprocessed/"
...
+# Compute the summary statistics based on sub-missions
+images_retained_by_sub_mission = compute_summary_statistics(
+  image_metadata,
+  "sub_mission_id",
+  metadata_per_sub_mission_filepath,
+  sub_mission_polygons_filepath
+)
+
+# Compute the images that were retained in both the mission polygons and the sub-mission polygons
+images_retained_in_both = intersect(images_retained_by_mission, images_retained_by_sub_mission)
```
youngdjn: Big logic comment/question: Shouldn't we use only this intersection set to compute the summary statistics, as opposed to computing them before as we are doing? If we throw out an image from a mission because it was thrown out from a sub-mission, I think we'd want the summary stats to reflect that. But I'm not sure the least repetitive way to do that since currently the image rejection and summary statistic computing happen together.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -27,93 +27,98 @@ EXIF_PATH = "/ofo-share/drone-imagery-organization/3b_exif-unprocessed/"
...
+  "mission_id",
+  metadata_per_mission_filepath,
+  mission_polygons_filepath
+)
+# Compute the summary statistics based on sub-missions
+images_retained_by_sub_mission = compute_summary_statistics(
+  image_metadata,
+  "sub_mission_id",
+  metadata_per_sub_mission_filepath,
+  sub_mission_polygons_filepath
```
youngdjn: Could you please name the function arguments?

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
```diff
@@ -3,29 +3,46 @@
...
-# Combine at the mission level
-metadata_mission = left_join(exif_mission, baserow_mission, by = c("dataset_id" = "dataset_id"))
-metadata_sub_mission = left_join(exif_sub_mission, baserow_sub_mission, by = c("dataset_id" = "dataset_id"))
-
+exif_metadata = read_csv(exif_metadata_filepath)
+
+# dataset_id is confusing so it is just dropped in general
+# In the context of a mission, the sub_mission_id has different meanings in baserow and the exif.
+# In baserow, it means the sub-missions included in a mission. In the exif it is the sub-mission
+# that that image is a part of.
```
youngdjn: Aren't these two the same thing? Since "the sub-mission that the image is a part of" is the "sub-mission included in a mission"?

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
```diff
@@ -3,29 +3,46 @@
...
+baserow_mission = subset(baserow_mission, select = -c(dataset_id, sub_mission_id))
+baserow_sub_mission = subset(baserow_sub_mission, select = -dataset_id)
```
youngdjn, line 33: Should we drop (or never create) these columns earlier or are they necessary earlier? If you don't know, I can keep an eye out for this when I do my revision.

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
```diff
@@ -3,29 +3,46 @@
...
+
+# There are name collisions for mission_id and sub_mission_id
```
youngdjn: Is this comment obsolete?

youngdjn (changes requested):
These are awesome improvements that really streamline this process and make it more comprehensible. Thanks! My comments are kinda terse but just because I was running out of time. Really happy with this overall.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: Yes, I agree that's a better way to approach it. This was just replicating the logic that was initially in here. I think if we factor out the boundary polygon identification step from the summary statistics it won't have to be redundant.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
russelldj: I moved both functions to the top. I think I initially didn't realize that R's variable scoping allowed functions at the top to use variables defined in the top-level scope, even if they had not been created at function definition.

sandbox/drone-imagery-ingestion/07_extract-exif-metadata-per-image.R
russelldj: I like this idea. I partially addressed this by excluding any sub-missions for which there wasn't a single baserow entry. However, the updated code would still break with an unsupported platform.

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
russelldj: Yes, removed

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
russelldj: I updated the comment to clarify. In one case, it is the list of sub-missions included in the mission that image is a part of. In the other case, it is the single sub-mission that the image is a part of.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: That was a mistake, but I'll save the renames for the end. I think a couple more files should be renamed so everything is sequential.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: I wasn't able to quite finish this, but I do have a start to it on the `dr_reorganize_ingest_wip` branch.

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
russelldj: I think we should be able to drop them but it will take a few more additional changes.

R/web-catalog-creation_drone-imagery-curation.R
russelldj: The `dataset_id` field is added [here](https://github.com/open-forest-observatory/ofo-r/blob/2ce5f187ffbf5e6c253507abd88902ae69d0064f/R/metadata-extraction_imagery_dataset.R#L647), which in turn assumes that it's in the per-image dataframe. I think that comment would be reasonable for now, but in the future we should remove the `dataset_id` from the summary entirely. Instead, we either hardcode this function to always use the `mission_id` as you described, or have a parameter for the `dataset_id_column` that can be set to either the `mission_id` or `sub_mission_id`.

russelldj:
Unfortunately I didn't get a chance to fully wrap all the nice-to-haves, but I think it could still be merged in the current state. I'm happy for you @youngdjn to make any changes you think are helpful before that.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: I tried moving it but then reverted it to being in the script. At this point I think it's pretty closely tied to the logic of the script so I'd leave it there, but I could go either way.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: This is now finished.

sandbox/drone-imagery-ingestion/11_merge-exif-and-baserow.R
russelldj: These were created in the `compile_baserow_metadata` step. I haven't looked much into that, so I'm not sure about removing them and would prefer to handle it on a different PR.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -27,93 +27,119 @@ EXIF_PATH = "/ofo-share/drone-imagery-organization/3b_exif-unprocessed/"
...
 exif_filepath = file.path(EXIF_PATH, paste0("exif_", IMAGERY_PROJECT_NAME, ".csv"))
 crosswalk_filepath = file.path(FOLDER_BASEROW_CROSSWALK_PATH, paste0(IMAGERY_PROJECT_NAME, "_crosswalk.csv"))
 
-metadata_perimage_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-exif-metadata_perimage_", IMAGERY_PROJECT_NAME, ".csv"))
-metadata_perdataset_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-exif-metadata_perdataset_", IMAGERY_PROJECT_NAME, ".csv"))
-polygons_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-polygons_", IMAGERY_PROJECT_NAME, ".gpkg"))
+# This per-image metadata was saved out in the last step
+metadata_perimage_input_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("exif-metadata_perimage_", IMAGERY_PROJECT_NAME, ".csv"))
+# Data for the subset of images retained in both the mission polygons and sub-mission polygons are saved out here
+metadata_perimage_subset_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("exif-metadata_perimage_subset_", IMAGERY_PROJECT_NAME, ".csv"))
```
russelldj, line 37: Consider renaming this to something more descriptive

R/web-catalog-creation_drone-imagery-curation.R
russelldj: I updated this to only expect `mission_id`. I think we can make it more general in the future if needed.

russelldj:
@youngdjn I think this is ready to go with the exception of renaming files, which I want to do after the final review. There's definitely more that could be done but the scope of this PR is getting pretty unwieldy so I'd like to get it merged and then make any subsequent changes later.

R/web-catalog-creation_drone-imagery-curation.R
youngdjn: Sounds good, looks good!

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn: Sounds good, we can move it later.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn: This is great!

youngdjn (approved):
This all looks good! Thanks for these additional changes. I agree we can merge now and potentially continue to remove any last vestiges of `dataset_id` that can be removed.

PR #59, opened 2025-01-22
Improve photogrammetry post-processing

russelldj:
This adds a couple of small changes.

- The ability to set bounds on which datasets should be converted
- The ability to control what fraction of the memory `terra` can access. The default 60% cap seems to lead to silent errors if gdal knows more memory will be required.
- Encourage the raster writer to use the bigtiff format by changing from the default flag of `IF_NEEDED` to `IF_SAFER`, see [here](https://gdal.org/en/stable/drivers/raster/gtiff.html). It seems that when compression is applied, bigtiff is not used even if the file is bigger than 4GB, at least when it is close to the limit.

### DIFF
--- a/sandbox/drone-imagery-ingestion/20_postprocess-photogrammetry-products.R
+++ b/sandbox/drone-imagery-ingestion/20_postprocess-photogrammetry-products.R
@@ -16,10 +16,12 @@ MISSION_FOOTPRINTS_PATH = "/ofo-share/drone-imagery-organization/3c_metadata-ext
 VISUALIZE = FALSE
 SKIP_EXISTING = FALSE
 
-RUN_CONVERSION = FALSE
+RUN_CONVERSION = TRUE
 RUN_CHM = TRUE
 RUN_THUMBNAIL = TRUE
 
+CONVERSION_LOWER_BOUND_DATASET = 1
+CONVERSION_UPPER_BOUND_DATASET = 10e+6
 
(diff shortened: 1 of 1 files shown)


sandbox/drone-imagery-ingestion/20_postprocess-photogrammetry-products.R
```diff
@@ -47,6 +60,12 @@ list_photogrammetry_outputs = function(input_folder) {
     timestamp = timestamps
   )
 
+  # Convert dataset_ids to indices and determine which ones are within the specified range of IDs
+  int_dataset_ids = strtoi(dataset_ids, base = 10)
+  elements_in_bounds = (int_dataset_ids >= lower_bounds_dataset) & (int_dataset_ids <= upper_bound_dataset)
```
youngdjn: I would surround the RHS by `which()` to remove `NA`s that could result here if the dataset ID coercion to integer produces any `NA`s (such as if there are letters in the dataset ID). `which` converts a vector of TRUE/FALSE/NA to the indices of the `TRUE` values only. Otherwise, indexing into a vector with a `NA` yields a `NA`.

youngdjn: And/or, return an error or warning if the coercion to integer yields `NA`s.

youngdjn (changes requested):
Good sleuthing! I didn't know about these tweaks. This looks great, just one minor suggestion to catch a possible future issue in case dataset IDs ever contain letters.

sandbox/drone-imagery-ingestion/20_postprocess-photogrammetry-products.R
russelldj: Thanks, I wasn't aware of that.

PR #60, opened 2025-01-23
Add patch for visualization of single ASL value

russelldj:
In the current version, the legend for ASL will have no information if there is only one distinct value. This is common for small datasets without terrain following since the ASL values are rounded to integers. In those cases, the categorical color mapping is applied so a label is still shown, just in a slightly different style.


youngdjn (approved):
Good fix!

PR #66, opened 2025-02-18
Rename drone-imagery-ingestion scripts to account for restructuring

russelldj:
This renames scripts in the `drone-imagery-ingestion` folder since some old scripts were removed or placed in a different order. The only re-ordering that happened in this script was moving the `metashape` config generation step later, since nothing currently depends on the outputs of it.


youngdjn (approved):
Looks good!

PR #68, opened 2025-03-06
Support missing `SerialNumber` metadata

russelldj:
Add an option to use the model name in place of the serial number if it's not present. This has shown up as a problem for the Matrice 100 platform.


youngdjn (approved):
Looks good!

PR #70, opened 2025-03-06
Replace dataset_id -> mission_id

russelldj:


youngdjn (commented):
This looks good! But one question: if we wanted to extend this function in the future to (alternatively) make a sub-mission details page, would this make it more difficult/confusing? Maybe this is one case where we should retain the `dataset_id` concept so that it can apply to either missions or sub-missions. But there may be a more important consideration behind this change that I'm missing.

russelldj:
Good point, I'll get rid of these changes and just add a line in the driver script to rename the column. I thought we properly handled this in the big overhaul PR, but it must have gotten lost in the shuffle.

russelldj:
Note that the logic in `web-catalog-cration_drone-imagery-curation` uses a mix of both `dataset_id` and `mission_id`. This should probably be updated in the future to just use one (dataset_id?).

youngdjn (approved):
Looks good!

Issue #42, opened 2024-12-14
Prevent mission polygon simplification from entirely eliminating small polygons


youngdjn:
In `extract_mission_polygon`, the simplification step can entirely eliminate small polygons. Possible solutions are:

- Reduce the degree of simplification
- Check if the polygon was entirely removed and if so, reduce the degree of simplification
- Apply a buffer before simplifying

youngdjn:
Closed by #77

Issue #61, opened 2025-02-04
In drone imagery ingest, use a constants file to define filenames (and possibly other constants) used across scripts


youngdjn:
So that we don't hard-code filenames (or even parts of filenames) in scripts, which could break the pipeline if we change it in one file but not another.

PR #77, opened 2025-03-19
Various pending ingest fixes and upgrades...

### LINKED ISSUE #67: In ingest step 08, consider improving conversion of list of polygons to single layer
Currently reliant on `do.call` and manual reassignment of CRS. Is there a more sf-native way to do this?

youngdjn:
See commit messages for items addressed.

This closes open-forest-observatory/ofo-r#67 and open-forest-observatory/ofo-r#42 .


sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -127,24 +146,40 @@ compute_and_save_summary_statistics = function(
...
+  image_metadata = image_metadata,
+  column_to_split_on = "mission_id",
+  image_merge_distance = IMAGE_MERGE_DISTANCE
+)
+sub_mission_res = compute_polygons_and_images_retained(
+  image_metadata = image_metadata,
+  column_to_split_on = "sub_mission_id",
+  image_merge_distance = IMAGE_MERGE_DISTANCE
+)
+
```
youngdjn, line 182: @russelldj , this is one realization I had: If we compute the mission footprints, then remove outlier images that were outside a sub-mission, this might mean we should then reduce the size of the mission polygon, right? So I added a second round of polygon computation. Does that make sense to you?

R/metadata-extraction_imagery_dataset.R
```diff
@@ -107,15 +107,38 @@ extract_mission_polygon = function(
...
+            "resulted in an empty polygon. No images will be retained.")
+  }
+
   # Identify which images are within the polygon if requested
   if (identify_images_in_polygon) {
     # increase the region to account for the simplified polygon
     simplified_poly_buffer = simplified_poly |> sf::st_buffer(simplification_tol)
-    intersection_idxs = sf::st_intersects(metadata, simplified_poly_buffer, sparse = FALSE)
+    intersection_mat = sf::st_intersects(metadata, simplified_poly_buffer, sparse = FALSE)
+    # Does the image intersect any of the polygons?
```
russelldj, line 134: Out of curiosity, what was the original behavior?

sandbox/drone-imagery-ingestion/03_compile-baserow-metadata.R
```diff
@@ -33,14 +33,32 @@ metadata_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-b
...
+  select(project_id, contributor_names, contact_info, license, objectives)
+baserow_datasets = baserow_datasets |>
+  left_join(baserow_projects, by = c("project_id" = "project_id")) |>
+  select(-project_id)
+
+# Apply any dataset-level contributor overrides
+baserow_datasets = baserow_datasets |>
+  mutate(contributor_names = ifelse(!is.na(contributor_names_override), contributor_names_override, contributor_names),
+         contact_info = ifelse(!is.na(contact_info_override), contact_info_override, contact_info)) |>
+  select(-contributor_names_override, contact_info_override)
```
russelldj: Is `contact_info_override` supposed to be dropped here as well?

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -3,16 +3,16 @@
...
-
 library(tidyverse)
 library(sf)
 
-# devtools::document(); devtools::install()
+# devtools::document()
+# devtools::install()
 library(ofo)
 
+IMAGE_MERGE_DISTANCE = 50
```
russelldj, line 13: Maybe in the future we could move all the major processing parameters to one `R` or `yml` file. Not necessary for this PR though.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
```diff
@@ -99,26 +124,20 @@ compute_and_save_summary_statistics = function(
...
-  # TODO in the future we might want to ensure that all the polygons have the same CRS but for now
-  # they should all be EPSG::4326
-  crs = sf::st_crs(polygons_perdataset[[1]])
-  polygons_perdataset_sf = sf::st_as_sf(polygons_perdataset_tbl, crs = crs)
+  # Transform the polygons into the appropriate format. The polygons begin as a named list, with the
+  # names corresponding to the dataset ID ("column_to_split_on"). The goal is to convert it to a sf
+  # object with each row representing a different dataset ID and the associated multi-polygon
+  # bounds.
+
+  polygons_perdataset_sf = dplyr::bind_rows(polygons_perdataset)
```
russelldj, line 132: Did I miss this simple fix or does it work now because of the new steps in `compute_polygons_and_images_retained`?

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
russelldj: Good catch, I think I've actually seen examples that may have been caused by this in the weird 557 dataset.

The one thing I'm trying to think through is whether this new polygon is guaranteed to include every image that is used to compute it. I think it's very likely, but I could see edge cases due to the simplification or the interaction between dropping images for the mission and sub-mission.

sandbox/drone-imagery-ingestion/10_merge-exif-and-baserow.R
```diff
@@ -14,16 +14,21 @@ IMAGERY_PROJECT_NAME = readr::read_lines(IMAGERY_PROJECT_NAME_FILE)
 EXTRACTED_METADATA_PATH = "/ofo-share/drone-imagery-organization/3c_metadata-extracted/"
 
 # Derived constants
+# In
 baserow_sub_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("sub-mission-baserow-metadata_", IMAGERY_PROJECT_NAME, ".csv"))
 baserow_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-baserow-metadata_", IMAGERY_PROJECT_NAME, ".csv"))
-exif_metadata_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("exif-metadata_perimage_subset_", IMAGERY_PROJECT_NAME, ".csv"))
+exif_metadata_mission_filepath = file.path(EXTRACTED_METADATA_PATH, paste0("mission-exif-metadata_perdataset_", IMAGERY_PROJECT_NAME, ".csv")) ### <- make sure this is based on the subset
```
russelldj: What does this comment mean?

russelldj (approved):
Looks good

russelldj:
> See commit messages for items addressed.
> 
> This closes open-forest-observatory/ofo-r#67 and open-forest-observatory/ofo-catalog-data-prep#2.

I don't think this is addressing open-forest-observatory/ofo-catalog-data-prep#2, which as I understand it would require additional logic in `extract_flight_terrain_correlation`. Is there something I missed?

youngdjn:
> I don't think this is addressing open-forest-observatory/ofo-catalog-data-prep#2, which as I understand it would require additional logic in `extract_flight_terrain_correlation`. Is there something I missed?

Oops good catch, I meant open-forest-observatory/ofo-r#42. I updated above.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn: Good point, I agree both of these may be possible, and actually could theoretically continue no matter how many times we iterated this. I think ultimately we need a more robust solution, probably integrated into a single function. I'm making an issue now. For now, I think any deviation will be minor and not lead to any significant errors downstream so I think it's OK to proceed with this unaddressed.

youngdjn: Added issue #80

R/metadata-extraction_imagery_dataset.R
youngdjn: It's that `st_intersects` returns a `n` x `m` boolean matrix for `n` image points x `m` polygons, not the indexes of images that are inside the polygons; a little more post-processing is necessary to get that.

sandbox/drone-imagery-ingestion/03_compile-baserow-metadata.R
youngdjn: Very good catch, this came back to bite me when trying to use this script since this edit. Correcting that.

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn: Totally agree. This should be addressed as part of #61 .

sandbox/drone-imagery-ingestion/08_extract-exif-metadata-mission.R
youngdjn: Yes, because of the new steps, specifically

` polygons_sf = purrr::map(polygons_sfc, st_as_sf)`

which illustrates a very tricky `sf` concept which is that geosptial objects without tabular attributes are technically `sfc` objects, and need to be converted to `sf` objects to carry tabular attributes. If you try to `bind_rows` of `sfc` objects, you get very unexpected behavior returning a data structure that is not compatible with any conventions that I'm aware of.

sandbox/drone-imagery-ingestion/10_merge-exif-and-baserow.R
youngdjn: Oops this was a note to self, removing.

PR #78, opened 2025-03-19
Improvements to imagery catalog creation

youngdjn:
Account for mission_id / sub_mission_id; format image count; add license and creator lookup

### DIFF
--- a/R/web-catalog-creation_drone-imagery-curation.R
+++ b/R/web-catalog-creation_drone-imagery-curation.R
@@ -1,16 +1,27 @@
 
 # Pull in all attributes to the plot summary table, compute relevant columns for display, (like
 # plot area in ha instead of m2), round numeric columns, create links to project and dataset pages
-compile_mission_summary_data = function(mission_level_metadata, base_ofo_url, mission_details_dir) {
+compile_mission_summary_data = function(mission_level_metadata, base_ofo_url, mission_details_dir, dataset_type = "mission") {
+  # dataset_type can be "mission" or "sub-mission"
 
-  # TODO: join relevant data to the mission metadata
+  # TODO: The dataset_id determination is flexible to either mission or sub-mission, but there is
+  # other hardcoded references to mission that would need to be updated if sub-mission is a
+  # possibility
 
-  # Select relevant columns to display
+  if (dataset_type == "mission") {
+    dataset_id_column = "mission_id"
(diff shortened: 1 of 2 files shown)


R/web-catalog-creation_drone-imagery-curation.R
```diff
@@ -1,23 +1,55 @@
...
-  # Select relevant columns to display
+  if (dataset_type == "mission") {
+    dataset_id_column = "mission_id"
+  } else if (dataset_type == "sub-mission") {
+    dataset_id_column = "sub_mission_id"
+  } else {
+    stop("dataset_type must be either 'mission' or 'sub-mission'")
+  }
+
+  mission_level_metadata$dataset_id = pull(mission_level_metadata, dataset_id_column)
```
russelldj, line 19: For the future, is `pull` preferred for extracting a column?

R/web-catalog-creation_drone-imagery-curation.R
```diff
@@ -1,23 +1,55 @@
...
+compile_mission_summary_data = function(mission_level_metadata, base_ofo_url, mission_details_dir, dataset_type = "mission") {
+  # dataset_type can be "mission" or "sub-mission"
 
-  # TODO: join relevant data to the mission metadata
+  # TODO: The dataset_id determination is flexible to either mission or sub-mission, but there is
+  # other hardcoded references to mission that would need to be updated if sub-mission is a
+  # possibility
 
-  # Select relevant columns to display
+  if (dataset_type == "mission") {
```
russelldj, line 11: Looks good. It seems like similar logic could be added in `make_mission_details_pages= and then all functions called by it could just use `dataset_id`.

russelldj (approved):

R/web-catalog-creation_drone-imagery-curation.R
youngdjn: In the case of `sf` objects, yes, because the base R equivalent,

`mission_level_metadata[, dataset_id_column]`

would return not just the requested column values, but the geometry column as well, and it would still be a `sf` object. It's an often-helpful but often-annoying and unintuitive behavior.

R/web-catalog-creation_drone-imagery-curation.R
youngdjn: Great, created as issue #81 .

PR #83, opened 2025-03-21
Fix some broken metadata extraction (file size, sub-mission ID list, comment typo)

youngdjn:
There were a few minor lingering issues of some metadata extraction that was apparently broken (or removed) by the switch from `dataset_id` to `mission_id`. This fixes/restores them.

- Raw imagery file size extraction
- Keep the list of `sub_mission_id`s that comprise a mission, but rename it to distinguish it from sub_mission_id as used in the raw imagery points metadata, which indicates which sub_mission the image is from.
- Retain project ID in mission-level metadata
- A comment typo


russelldj:
Sorry I missed this

PR #90, opened 2025-08-12
Refactor ITD for S3 bucket

youngdjn:
Retrieves CHM file listing from S3, and runs the task of downloading the CHM, computing the treetops and crowns, uploading the products, and deleting the local files, in parallel for each mission that has a CHM.


youngdjn:
Accidental destination of the PR

PR #91, opened 2025-09-18
Dy/bugfix/misaligned image ids in exif

youngdjn:
The EXIF parsing function included some reordering (by collection time) and subsetting (excluding images without GPS data), which messed up the correspondence between the raw EXIF (which is later used to populate some additional attributes of the parsed EXIF) and the parsed EXIF, which in turn led to the wrong images being indexed in some cases. This fix ensures the data frame returned by the exif parsing function is the same order and length as the data frame passed to it.


youngdjn:
Accidental PR to old repo

